PR #17, opened 2019-03-25
FIX Do not call remote for cuml repo

mike-wendt:


mike-wendt:
This should be implemented as a flag passed to the clone, instead of hard coded

Closing and will open issue to address this better

Issue #24, opened 2019-06-17
[FEA] Remove cuda labels from conda channels


rlratzel:
**Is your feature request related to a problem? Please describe.**
This change keeps consistency between the Dockerfiles generated here and the conda installs done in gpuCI.

**Describe the solution you'd like**
Channel names consistent with those in the `.condarc` shown in this PR: https://github.com/rapidsai/gpuci-build-environment/pull/25

**Additional context**
https://github.com/rapidsai/gpuci-build-environment/pull/25

mluukkainen:
WIP patch here: https://github.com/rapidsai/build/pull/27

mike-wendt:
Given that we can build and use the containers with the changes in `branch-0.9` I think we should put this issue and #27 on the backlog and revisit after the release

mluukkainen:
I moved this against branch-0.10 and images build. From my POV it could go in if the PR is still needed:

https://github.com/rapidsai/build/pull/27/files

PR #27, opened 2019-07-31
Remove cuda labels from .condarc

mluukkainen:
edit. gpuci dockerfile seems to have been removed. No problem anymore with it.

### DIFF
--- a/supportfiles/.condarc
+++ b/supportfiles/.condarc
@@ -1,0 +1,2 @@
+channels:
+  - conda-forge
--- a/commands/buildDockerImage.sh
+++ b/commands/buildDockerImage.sh
@@ -113,10 +113,10 @@ mkdir -p ${RAPIDS_SOURCES_DIR}
 
 # Copy other dirs that many Dockerfiles expect to copy from their CWD
 # FIXME: ensure the copy succeeds
-# FIXME: copy the .condarc files directly to the OUTPUT_DIR since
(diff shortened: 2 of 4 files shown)


rlratzel:
> Not sure what to do with file: generatedDockerfiles/Dockerfile.centos7-gpuci

I would just apply the same changes to it and resolve the conflicts.

mluukkainen:
cugraph build on branch-0.9 is broken. I see this error when building clean branch-0.9:
cugraph/utilities/column_utils.pyx:60:36: undeclared name not builtin: get_dtype

cugraph build error blocks building docker images.

mluukkainen:
There is so many errors when running test.sh using new containers built with branch-0.9 that I'm unable to determine if this patch is working as intended. Blocked on getting this merged/reviewed due to broken tests.

mike-wendt:
As mentioned in #24 lets put this on the backlog as the code changes in `branch-0.9` work without the need for the `.condarc` file. We can troubleshoot this more after the release

supportfiles/.condarc
```diff
@@ -0,0 +1,3 @@
+channels:
+  - conda-forge
+  - defaults
```
rlratzel: I think we want to remove `defaults` since it usually causes problems with the solver and other things (sorry I don't have specifics).  Can you remove and re-test?

PR #33, opened 2019-09-09
Fix runtime container building error

mluukkainen:


rlratzel:
I'm assuming that this will need to be updated in the gpuci images as a longer-term fix. I can open a separate issue/PR for that.

PR #36, opened 2019-09-10
Add cuspatial

mluukkainen:
- Add cuspatial to base and devel images. Runtime image inherits from base
- Move RAPIDS_CONDA_VERSION_SPEC to 0.10* due to cuspatial dependencies
- Regenerate Dockerfiles


generatedDockerfiles/Dockerfile.centos7-devel
```diff
@@ -212,6 +217,7 @@ CMD [ "/bin/bash" ]
 RUN source activate rapids \
     && env \
     && conda list \
+    && conda uninstall -y blas openblas \
```
mluukkainen: Blas is coming from previous commit. It looks like dockerfiles were not regenerated for blas patch and they show up now

dockerArgs
```diff
@@ -15,7 +15,7 @@ ARG IPYTHON_VERSION=7.3*
 
 # base/runtime install specs
 ARG DASK_XGBOOST_CONDA_VERSION_SPEC=0.2*
-ARG RAPIDS_CONDA_VERSION_SPEC=0.9*
+ARG RAPIDS_CONDA_VERSION_SPEC=0.10*
```
mluukkainen: cuspatial was not available for 0.09 so I had to bump version to 0.10* to make adding cuspatial possible

generatedDockerfiles/Dockerfile.centos7-devel
```diff
@@ -180,6 +180,11 @@ RUN cd ${RAPIDS_DIR}/rmm && \
 RUN cd ${RAPIDS_DIR}/cudf && \
     source activate rapids && \
     ./build.sh
+RUN cd ${RAPIDS_DIR}/cuspatial && \
+    source activate rapids && \
+    export CUSPATIAL_HOME="$PWD" && \
```
mluukkainen: Should I do export here or maybe use the generated build script similar to xgboost?

rlratzel: This is fine for now. Can you also put a FIXME comment saying that we shouldn't need to set these vars long-term?

generatedDockerfiles/Dockerfile.centos7-devel
rlratzel: Hopefully they'll get updated. Can you put a FIXME comment saying this should also be temporary?

dockerArgs
rlratzel: This seems correct to me, and probably something we needed to do earlier anyway.

generatedDockerfiles/Dockerfile.centos7-devel
mluukkainen: Fixed

generatedDockerfiles/Dockerfile.centos7-devel
mluukkainen: Fixed

generatedDockerfiles/Dockerfile.centos7-devel
```diff
@@ -147,7 +147,7 @@ RUN source activate rapids \
     && conda remove --force \
          rmm librmm \
          nvstrings libnvstrings \
-         cudf dask-cudf libcudf \
+         cudf cuspatial dask-cudf libcudf \
```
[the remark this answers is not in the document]
mluukkainen: Fixed

PR #54, opened 2020-02-04
Note cuxfilter is only in runtime containers

mike-wendt:
Due to it's dependencies it is not in the `base` or `devel` container

### DIFF
--- a/doc/NGC_rapidsai.md
+++ b/doc/NGC_rapidsai.md
@@ -10,19 +10,20 @@ The RAPIDS suite of software libraries gives you the freedom to execute end-to-e
 
 ### Current Version
 
+#### RAPIDS 0.12 - 4 February 2020
+
+Versions of libraries included in the `0.12` images:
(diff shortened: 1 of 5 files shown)


doc/NGC_rapidsai.md
```diff
@@ -31,10 +32,12 @@ images in order to make it easy to add RAPIDS libraries while maintaining suppor
...
+- `base` - contains a RAPIDS environment ready for use.
+  - **TIP: Use this image if you want to use RAPIDS as a part of your pipeline.**
+- `runtime` - extends the `base` image by adding a notebook server and example notebooks.
+  - **TIP: Use this image if you want to explore RAPIDS through notebooks and examples.**
 
-For images that contain the full RAPIDS source tree, pre-built with all artifacts in place, and the compiler toolchain, the debugging tools, the headers and the static libraries for RAPIDS development - as well as support for Python 3.7 - refer to the [rapidsai/rapidsai-dev](https://cloud.docker.com/u/rapidsai/repository/docker/rapidsai/rapidsai-dev) repo on DockerHub.
+For images that contain the full RAPIDS source tree, pre-built with all artifacts in place, and the compiler toolchain, the debugging tools, the headers and the static libraries for RAPIDS development - as well as support for Python 3.7 - refer to  [rapidsai/rapidsai](https://cloud.docker.com/u/rapidsai/repository/docker/rapidsai/rapidsai) and [rapidsai/rapidsai-dev](https://cloud.docker.com/u/rapidsai/repository/docker/rapidsai/rapidsai-dev) repos on DockerHub.
```
rlratzel: Adding `rapidsai/rapidsai` might be misleading since this paragraph is specifically describing `devel` images. Maybe we need something saying:
"For `base` and `runtime` images with python 3.7 support, refer to [rapidsai/rapidsai] on DockerHub.  For images that contain the full RAPIDS source tree...."

mike-wendt: I agree I'll split this out to remove any confusion

mike-wendt:
Merging and deploying. Can update if needed.

PR #56, opened 2020-02-14
Use new conda rapids env meta-pkgs

mike-wendt:
Convert to using new meta-pkgs that setup build, doc, and notebook environments. This consolidates all definitions in one conda config and allows them to be reused in gpuCI and repos for their testing with one place to change them.


mike-wendt:
> I noticed that the dependencies removed from this PR don't map to the three new conda recipes 1:1, but everything built okay for me so I'm assuming that's fine.

Yes these also include build dependencies that we didn’t include previously but are needed for testing or a part of the conda build requirements across the repos. Also for consistency some repo wide dependencies that were in the docker container below these were rolled up so we change them in one environment across the repo.

This “base” layers from the `gpuci` Docker hub repo also need updating to remove some of the redundant installs here. That is a separate effort but I wanted to get this out first.

PR #63, opened 2020-03-12
Update permissions on quick template

ajschmidt8:
This PR updates the permissions in `/opt/conda` for `quick` builds. This is needed since the `quick` builds perform conda operations which modifiy the permissions in `/opt/conda` subdirectories.


ajschmidt8:
before changes:

image


after changes:

image

PR #64, opened 2020-03-13
[WIP] Update notebook testing and cudf legacy test build option

dillon-cullinan:
Merge when runtime builds with new notebooks repo are green.

The `-l` addition to the cudf build script is for newly added optional legacy tests.

### DIFF
--- a/supportfiles/test.sh
+++ b/supportfiles/test.sh
@@ -36,7 +36,7 @@ for nb in $(find repos/*/notebooks/* -name *.ipynb); do
         ${NBTEST} ${nbBasename}
         EXITCODE=$((EXITCODE | $?))
         rm -rf ${LIBCUDF_KERNEL_CACHE_PATH}/*
-        cd ${RAPIDS_DIR}/${folder}
+        cd ${RAPIDS_DIR}
     fi
 done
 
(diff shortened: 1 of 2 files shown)


supportfiles/test.sh
```diff
@@ -20,32 +18,26 @@ EXITCODE=0
...
-        else
-            cd $(dirname ${nb})
-            nvidia-smi
-            ${NBTEST} ${nbBasename}
-            EXITCODE=$((EXITCODE | $?))
-            rm -rf ${LIBCUDF_KERNEL_CACHE_PATH}/*
-            cd ${NOTEBOOKS_DIR}/${folder}
-        fi
-    done
+for nb in $(find repos/*/notebooks/* -name *.ipynb); do
```
rlratzel: minor: the old code used a var to make it a little more self-documenting, but if you don't want to use a var you might consider adding a comment to explain to future maintainers what this `find` call is intended to find.

rlratzel: I was actually thinking of the old code that did this: `for folder in ${TOPLEVEL_NB_FOLDERS};`, but I see your new code just uses one `find` to find all NBs under the `repos` dir.  A comment might be nice but not critical - totally up to you.

PR #65, opened 2020-03-18
add conda retry interval -> 30s

ajschmidt8:
This PR increases the default conda retry interval to 30s to account for the amount of conda errors we've been seeing in base builds lately.


ajschmidt8:
closing this since mike found the root cause of the issue in this thread

https://nvidia.slack.com/archives/CJWBS14NB/p1584539865246300

PR #66, opened 2020-03-25
Fix gcc 7 update issues

mike-wendt:
This fixes a problem for Ubuntu 18.04 where gcc was being updated to 7.5 from the `nvidia/cuda` version of 7.4. With 7.5 there are conflicts with GDAL and CUDA in cuspatial. Instead of updating our version of gcc in Ubuntu we are now relying on the version installed by `nvidia-cuda` containers

### DIFF
--- a/templates/docker/gcc7_centos7
+++ b/templates/docker/gcc7_centos7
@@ -1,26 +1,3 @@
-##TODO-mwendt I don't think we need NCCL installed and most of these pkgs are installed
-###
-### The support dir contains RPMs that enable additional repos needed
-### for CentOS (among other things). Copy them to a temp dir and remove
-### after installed.
-###
-##COPY supportfiles/*.rpm /tmp
-##
-##RUN yum install -y /tmp/*.rpm && \
-##    rm -rf /tmp/*.rpm && \
-##    yum upgrade -y && \
-##    yum install -y \
-##      bzip2 \
-##      curl \
-##      git \
-##      screen \
-##      vim \
-##      wget \
-##      which \
-##      libnccl-2.4.2-1+cuda${CUDA_MAJORMINOR_VERSION} \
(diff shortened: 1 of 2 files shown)


templates/docker/conda_install_devel_deps
```diff
@@ -9,7 +9,9 @@ RUN source activate rapids \
 
 # Install rapids-build-env from conda meta-pkg
 RUN ${RAPIDS_DIR}/utils/condaretry install -y -n rapids --freeze-installed \
-        rapids-build-env=${RAPIDS_CONDA_VERSION_SPEC}
+        rapids-build-env=${RAPIDS_CONDA_VERSION_SPEC} \
+    && conda remove -y -n rapids --force rapids-build-env \
```
ajschmidt8: I assume we're uninstalling `rapids-build-env` because it includes an earlier version of numba ([src](https://github.com/rapidsai/integration/blob/branch-0.13/conda/recipes/rapids-build-env/meta.yaml#L71)). will we be updating `numba` in `rapids-build-env`? if so, when? also, what about all the other packages from `rapids-build-env`? do they need to be re-installed? it looks like now we're only reinstalling numba below.

mike-wendt: I'm working on the updates to the integration repo now to publish a new meta-pkg with numba 0.48. This commit will be reverted once I have the new meta-pkgs built.

mike-wendt: This is still a WIP PR, I copied the labels over but forgot to add it

mike-wendt:
The update to numba 0.48 is done through the `rapids` and `rapids-build-env` packages that were updated in rapidsai/integration#15

mike-wendt:
This also fixes a problem for Ubuntu 18.04 where gcc was being updated to 7.5 from the `nvidia/cuda` version of 7.4

PR #71, opened 2020-03-30
[REVIEW] Properly output repo name and notebook name on junit output

dillon-cullinan:


supportfiles/test.sh
```diff
@@ -23,6 +23,14 @@ EXITCODE=0
...
     nbBasename=$(basename ${nb})
+    # Output of find command looks like this: ./repos/<repo>/notebooks/<notebook> -name
+    # This grabs the <repo> element
+    nbRepo=$(echo ${nb} | awk -F/ '{print $3}')
+
+    echo "========================================"
+    echo "REPO: ${nbRepo}"
+    echo "========================================"
+
     # Skip all NBs that use dask (in the code or even in their name)
```
rlratzel: minor: might want to update the comment to say you're also skipping CLK NBs and why.  (I should have included why we're skipping dask, which is because we aren't currently automating the required startup of a separate dask server in order for those NBs to run - without the server, they error out).

PR #75, opened 2020-04-21
[REVIEW] Convert notebooks into working directory

dillon-cullinan:
Some of our repos use datasets in their notebooks which are referenced using relative paths. This change makes the python scripts generated from `nbconvert` to be placed into the working directory. This guarantees that any relative paths stay consistent.

In the test script that calls this, there is already a `cd` command into the notebook's directory before running, which will avoid any issues involving working directory. https://github.com/rapidsai/build/blob/branch-0.14/supportfiles/test.sh#L49

### DIFF
--- a/utils/nbtest.sh
+++ b/utils/nbtest.sh
@@ -28,13 +28,13 @@ for nb in $*; do
     NBFILENAME=$1
     NBNAME=${NBFILENAME%.*}
     NBNAME=${NBNAME##*/}
-    NBTESTSCRIPT=/${NBDIR}/${NBNAME}-test.py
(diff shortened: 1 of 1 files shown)


rlratzel:
A notebook that uses relative paths still needs to be written wrt where the script is **run** (ie. the `CWD`), not where the temp file is located.  IOW, writing to `/tmp` shouldn't be a problem if the script is run from the correct location (and the script itself doesn't `cd` inappropriately).  These scripts have been working as-is for many months, so do we know what changed to cause things to break? I'm concerned that dumping all the temp files into CWD will make a mess and potentially add confusion.

PR #81, opened 2020-05-19
Add GitHub Action for Generating Dockerfiles

ajschmidt8:
This PR adds a GitHub action to automatically generate the latest Dockerfiles anytime a PR is merged to the repo's default branch.

### DIFF
--- a/.github/workflows/dockerfiles.yaml
+++ b/.github/workflows/dockerfiles.yaml
@@ -9,22 +9,24 @@ jobs:
     name: Update Dockerfiles on Default Branch Merge
     runs-on: ubuntu-latest
     container: python:3.8
-    if: github.event.pull_request.base.ref == github.event.repository.default_branch
+    if: github.event.pull_request.base.ref == github.event.repository.default_branch && github.event.pull_request.merged == true
     steps:
       - name: Checkout repo
         uses: actions/checkout@v2.1.1
       - name: Generate new Dockerfiles
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/dockerfiles.yaml
```diff
@@ -0,0 +1,30 @@
+name: Update Dockerfiles on Default Branch Merge
+
+on:
+  pull_request:
+    types: [closed]
```
dillon-cullinan, line 5: Is there an option for merged only?

.github/workflows/dockerfiles.yaml
```diff
@@ -0,0 +1,30 @@
...
+        python generate_dockerfiles.py
+        mv build/* generated-dockerfiles/
+      - name: Setup git Config
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
+      - name: Commit to Repository
+        run: |
+          git add generated-dockerfiles/*
+          git commit -m 'update dockerfiles'
```
dillon-cullinan, line 29: Automated commits should have a more descriptive response. Maybe something like

`[GH Actions] Generate Dockerfiles for PR #XX`

.github/workflows/dockerfiles.yaml
ajschmidt8: yes, good catch. there's a `merged` key in the webhook payload that I'll need to check for that.

PR #101, opened 2020-06-09
Remove nvstrings build step

ajschmidt8:
This PR removes the `build_tests_nvstrings` build step from our repo. It looks like it was removed from `cudf` in PR https://github.com/rapidsai/cudf/pull/5373 and is now causing issues in our builds.

### DIFF
--- a/templates/partials/devel_build.dockerfile.j2
+++ b/templates/partials/devel_build.dockerfile.j2
@@ -10,7 +10,7 @@ RUN cd ${RAPIDS_DIR}/{{ lib.name }} && \
   {% if lib.name == "cudf" %}
   ./build.sh -l && \
   cd cpp/build && \
-  make -j${PARALLEL_LEVEL} build_tests_cudf
+  ./build.sh tests
 
   {% elif lib.name == "cuspatial" %}
   export CUSPATIAL_HOME="$PWD" && \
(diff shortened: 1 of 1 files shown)


templates/partials/devel_build.dockerfile.j2
```diff
@@ -10,7 +10,6 @@ RUN cd ${RAPIDS_DIR}/{{ lib.name }} && \
   {% if lib.name == "cudf" %}
   ./build.sh -l && \
   cd cpp/build && \
-  make -j${PARALLEL_LEVEL} build_tests_nvstrings && \
   make -j${PARALLEL_LEVEL} build_tests_cudf
```
rlratzel, line 13: I'd replace this with `./build.sh tests` so they don't break us in the future by changing target names, etc.

rlratzel, line 13: I'd replace this with ./build.sh tests so they don't break us in the future by changing target names, etc.

ajschmidt8:
> sorry I wasn't clear, since `build.sh` is in the root, you should not be doing the `cd` call above.
> 
> Also, you _should_ be able to combine build.sh targets/options like so: `./build.sh -l tests` especially in this case since `-l` and `tests` are both options for building tests. That's optional though if you want it cleaner, and I'd test it first.
> 
> disclaimer: I haven't tested this locally, just suggesting based on code inspection here

whoops. got it. also dropped the `-l` based on this pr https://github.com/rapidsai/cudf/pull/5430

PR #113, opened 2020-07-06
FIX Install RAPIDS in `runtime` containers

mike-wendt:
Previously `runtime` containers FROM-ed `base` where RAPIDS was already installed. The changes made to the build process broke that dependency so we have to install RAPIDS in `runtime` separately.

This is related to an issue that was reported where `import cuml` does not work in `runtime` images. The output from `conda list` does not show `cuml` or any RAPIDS packages are installed with the current `runtime` images. The change in this PR installs RAPIDS to fix this issue.


aleksficek:
LGTM!

PR #118, opened 2020-07-07
[REVIEW] Initial version of devel images with BlazingSQL added

rlratzel:
This also include the "Welcome to BlazingSQL" notebook repo and updates to `/test.sh` to run the welcome NB.

Note: There are other NBs in the Welcome repo, but those are not being run since they require additional dependencies. When the `runtime` images are using the updated RAPIDS meta-packages (specifically an updated notebook deps package), this can be revisited if it's decided the additional dependencies for the extra Blazing NBs are to be included.

@ajschmidt8 - I left the "clean" option in the updated generate script if you ever find that useful. I can remove that if you don't want that code in there.


templates/partials/clone_and_build_blazing.j2
```diff
@@ -0,0 +1,27 @@
+{# This partial clones the BlazingSQL repo and build and installs it into the rapids conda env. #}
+
+{# Install build prerequisites #}
+RUN source activate rapids \
+    && conda install -y \
```
ajschmidt8, line 5: do you want to use `gpuci_conda_retry` here?

rlratzel: Yes. I think I hit the same issue we saw in the benchmarking job where it wasn't using the right env, but I'll work on it since it should be using `gpuci_conda_retry`

generate_dockerfiles.py
```diff
@@ -27,17 +27,14 @@ def load_settings():
 
 
 def initialize_output_dir():
-    """Creates or empties the OUTPUT_DIR directory"""
+    """Creates or empties (if clean==True) the OUTPUT_DIR directory"""
```
ajschmidt8, line 30:
```suggestion
    """Creates the OUTPUT_DIR directory"""
```

rlratzel:
@mike-wendt there's a long internal email thread that ended with "let’s get the devel containers good to go while we work on improving the conda packaging situation", which I interpreted to be what's in this PR. I didn't realize the original request was still what we wanted after that. I can make the change you're asking for, but like you said, it'll now have to wait for 0.16.

datametrician:
Let's ask clarifying questions more, and get sign off. This mistake only cost BlazingSQL a few dozen thousand downloads :(

rlratzel:
> Let's ask clarifying questions more, and get sign off.

Agreed. Unfortunately, clarification and sign off was the intent of the PR :(  but we'll figure out a different way.

ajschmidt8:
closing this in favor of #179

PR #119, opened 2020-07-07
FIX Correct Dockerfile selection in gpuCI script

mike-wendt:
The previous selector chose the `centos7` Dockerfiles for `ubuntu16.04` and `ubuntu18.04`. This resulted in errors with the build that prevented the ENTRYPOINT to work correctly and despite local builds working, the produced builds from gpuCI did not have the correct `/rapids` directory and the necessary scripts/notebooks as expected.

I tested these changes in the gpuCI script job and confirmed they work as expected.


mike-wendt:
Successful build & test with https://gpuci.gpuopenanalytics.com/job/rapidsai/job/docker/job/rapidsai-base-runtime/67/ will merge so overnight builds use this

PR #121, opened 2020-07-07
[REVIEW] Add Ccache to Devel Images

aleksficek:
Closes https://github.com/rapidsai/ops/issues/995

Based on the potential to break other builds (as seen last Wednesday https://github.com/rapidsai/gpuci-build-environment/pull/102) and difficulty in merging two PR's from two separate repo's, both building ccache from source and utilizing it will be in the `rapidsai/rapidsai-nightly` devel images.

**Necessary Steps After Merge**
- [ ] Segment devel jobs into two (1 for uploading to S3 and other for downloading from S3)
- [ ] Add corresponding scripts to pull/push from S3


templates/Devel.dockerfile.j2
```diff
@@ -15,6 +15,35 @@ ARG FROM_IMAGE=gpuci/rapidsai
...
+ARG CCACHE_VERSION=master
+
+{# Install ccache from source (use specific version in projects commit history) #}
+RUN git clone https://github.com/ccache/ccache.git /tmp/ccache && cd /tmp/ccache \
+ && git checkout -b rapids-compose-tmp b1fcfbca224b2af5b6499794edd8615dbc3dc7b5 \
+ && ./autogen.sh \
+ && ./configure --disable-man --with-libb2-from-internet --with-libzstd-from-internet\
+ && make install -j \
+ && cd / \
+ && rm -rf /tmp/ccache-${CCACHE_VERSION}* \
```
mike-wendt, line 27: Looks like from line 21 you're installing to `/tmp/ccache` and checking out a specific commit on line 22. Doesn't look like the ARG on line 18 is used and the remove on line 27 appears that it would be leaving the install files given that you're using `/tmp/ccache`

May want to drop line 18, and change line 27 to match `/tmp/ccache`

templates/Devel.dockerfile.j2
```diff
@@ -86,6 +115,12 @@ numba.cuda cannot load that, and instead have it load
 ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH_PREBUILD}
 {% endif %}
 
+{# Report ccache stats and cleanup; also fix  #}
+RUN ccache -s
+RUN ccache -c \
+    && chmod -R ugo+w /ccache
+RUN ccache -s
```
mike-wendt, line 122: I would combine these into a single run and drop the `chmod` command here. If needed you can add echos, for context info.

Also I would move `/ccache` to the `chmod` command in `partials/cleanup-chmod.dockerfile.j2`. This way it is done all at once

templates/partials/cleanup-chmod.dockerfile.j2
```diff
@@ -2,4 +2,4 @@
 
 {# Clean-up conda and update ACLs for all users #}
 RUN conda clean -afy \
-  && chmod -R ugo+w /opt/conda ${RAPIDS_DIR}
\ No newline at end of file
+  && chmod -R ugo+w /opt/conda ${RAPIDS_DIR}
```
mike-wendt, line 5: As mentioned in the `Devel` template, I would add `/ccache` here

templates/partials/devel_build.dockerfile.j2
```diff
@@ -10,6 +10,8 @@ ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/opt/conda/envs/rapids/lib
 {% for lib in RAPIDS_LIBS %}
 RUN cd ${RAPIDS_DIR}/{{ lib.name }} && \
   source activate rapids && \
+  env && \
```
mike-wendt, line 13: Do we need to keep `env` for each build? It does add an extra amount of logging, would moving this up out of the loop be sufficient?

templates/Devel.dockerfile.j2
aleksficek: I combined the above into a single command and I hear the sentiment about having all the chmod cleanup together. The issue with moving the `chmod /ccache` is that the command will also be distributed to non-devel images which don't use ccache. Do you think this will be a problem?

templates/partials/devel_build.dockerfile.j2
aleksficek: Removed `env` and `ccache` as they are unnecessary. The `ccache` command might also cause some issues in the `quick` images if they aren't updated to FROM the new devel images in time.

templates/Devel.dockerfile.j2
mike-wendt: Fair point, I think there is a way to leverage the template-ing language to do an if statement, but what you have now if good enough

PR #123, opened 2020-07-09
[REVIEW] Remove Python 3.6 and CUDA 10.0

aleksficek:
Moving away from Python 3.6 and CUDA 10.0 as seen in https://github.com/rapidsai/ops/issues/1017

The following are grey areas as to whether these should be changed:
- [x] Legacy Docker CE v18 https://github.com/rapidsai/docker/blob/drop-python-cuda/dockerhub-readme/Dockerhub_rapidsai-dev.md#legacy---docker-ce-v18-and-nvidia-docker2-1 and https://github.com/rapidsai/docker/blob/dfb8706f9555427110b1fc1207a1e3fe22d1c2f2/dockerhub-readme/Dockerhub_rapidsai.md#legacy---docker-ce-v18-and-nvidia-docker2


CONTRIBUTING.md
```diff
@@ -17,7 +17,7 @@ This repo uses [Jinja2](https://www.palletsprojects.com/p/jinja/) to generate Do
 
 Make sure to have the following dependencies installed to compile the Dockerfiles.
 
-- Python 3.6+
+- Python 3.7+
```
ajschmidt8, line 20: This Python version should stay as `3.6+`. This indicates the version of Python that is required to run `generate_dockerfiles.py`. It's independent of the Python/CUDA versions used for the images themselves.

PR #124, opened 2020-07-09
[REVIEW] Add Notebook Ports Instructions

### LINKED ISSUE #120: [ENH] Add info message about Jupyter notebook server to ENTRYPOINT
https://github.com/rapidsai/docker/blob/23027b8b7b64d6f4f7441956b4facf6f3273bf46/context/.start_jupyter_run_in_rapids.sh#L3

After the line above, I think we should add some instructions for the user to let them know the notebook server has been started and how to access it. Let me know your thoughts on this.

aleksficek:
Closes #120

Added output of what ports need to be exposed when running the runtime/devel container.

### DIFF
--- a/context/.start_jupyter_run_in_rapids.sh
+++ b/context/.start_jupyter_run_in_rapids.sh
@@ -3,4 +3,6 @@ source activate rapids
 /rapids/utils/start_jupyter.sh > /dev/null 
 echo "Notebook server successfully started!"
 echo "To access visit http://localhost:8888 on your host machine."
+echo 'Ensure the following args to "docker run" are added to expose the server ports to your host machine:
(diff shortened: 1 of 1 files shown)


context/.start_jupyter_run_in_rapids.sh
```diff
@@ -3,4 +3,6 @@ source activate rapids
 /rapids/utils/start_jupyter.sh > /dev/null 
 echo "Notebook server successfully started!"
 echo "To access visit http://localhost:8888 on your host machine."
+echo 'Ensure the following args to "docker run" are added to expose the server ports to your host machine:
```
ajschmidt8, line 6: can you change `args` to `arguments`?

PR #129, opened 2020-07-24
[REVIEW] Temporary Devel Ccache Fix

aleksficek:
With the update to ccache 3.7.11 the cache has been bloating up to 4.2 GB and not performing well. I have found a range of commit's in ccache's master branch that both solve the cmake `Cannot build a simple test program`  error experienced last week and keep ccache performing as it previously did. The PR installs ccache at a commit in this Goldilocks Zone.

This is a temporary fix and early next week this code will be replaced with the work in progress conda based install (https://github.com/rapidsai/ccache-feedstock/tree/enh-rapids-ccache)


aleksficek:
Made quick progress as part of https://github.com/rapidsai/ccache-feedstock/tree/enh-rapids-ccache so the temporary fix has lost it's value. Closing this PR.

PR #130, opened 2020-07-29
[REVIEW] Install ccache with forked ccache-feedstock package

aleksficek:
Builds off of the recent changes in our forked version of the ccache feedstock https://github.com/rapidsai/ccache-feedstock/pull/1

The version of ccache in the `gpuci-ccache` package is a a specific commit in history which does not bloat ccache and does not not cause the cmake `cannot compile a simple test program` error.

### DIFF
--- a/templates/Devel.dockerfile.j2
+++ b/templates/Devel.dockerfile.j2
@@ -23,7 +23,7 @@ RUN source activate rapids \
 RUN conda install conda-build
 RUN echo "$(pwd)"
 RUN conda build .
-RUN conda install --use-local ccache
+RUN conda install -c conda-forge --use-local ccache
 
 
 ENV CCACHE_NOHASHDIR=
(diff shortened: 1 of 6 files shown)


templates/Devel.dockerfile.j2
```diff
@@ -15,11 +15,9 @@ ARG FROM_IMAGE=gpuci/rapidsai
...
 FROM ${FROM_IMAGE}:${RAPIDS_VER}-cuda${CUDA_VER}-devel-${LINUX_VER}-py${PYTHON_VER}
 
-{# Install ccache from source (use specific version in projects commit history) #}
-ARG CCACHE_VERSION=3.7.11
-RUN curl -s -L https://github.com/ccache/ccache/releases/download/v${CCACHE_VERSION}/ccache-${CCACHE_VERSION}.tar.gz -o ccache-${CCACHE_VERSION}.tar.gz \
- && tar -xvzf ccache-${CCACHE_VERSION}.tar.gz && cd ccache-${CCACHE_VERSION} \
- && ./configure && make install -j16 && cd - && rm -rf ./ccache-${CCACHE_VERSION} ./ccache-${CCACHE_VERSION}.tar.gz
+{# Install ccache from conda package #}
+RUN source activate rapids
+RUN conda install -c gpuci gpuci-ccache
```
ajschmidt8, line 20:
```suggestion
RUN source activate rapids && \
    conda install -c gpuci gpuci-ccache
```

I think these two lines need to be consolidated in order for the `rapids` environment to be utilized during the `conda install`. Feel free to double-check this though.

aleksficek: Good catch, ccache should be in the base environment anyways so I just removed the `source activate rapids` part.

PR #132, opened 2020-08-03
[REVIEW] Add Container Entrypoint Info

### LINKED ISSUE #96: [FEA] Add info message to container entrypoint
From @miguelangel :

> Just a quick observation about changes from 0.8 to 0.9 container.
More specifically, I have noticed that, from 0.9 container, a JupyterLab instance is executed when docker run the container.
That is very handy, but as a side effect, it took me a while to understand why local mapped folder was not visible anymore.
As a quick workaround, I have mapped my folder to /rapids/notebook, to make it visible to JupyterLab.
>> docker run --runtime=nvidia --rm -it -v /raid/miguelm:/rapids/notebooks/miguelm  -p 8888:8888 -p 8787:8787 -p 8786:8786  nvcr.io/nvidia/rapidsai/rapidsai:0.9-cuda10.0-runtime-ubuntu18.04

> Might I suggest to explicitly mention the above somewhere? Apologies if it is already detailed somewhere, but I couldn't find it.

> What about an echo message when running the container?

@rmccorm4 added:

> ...NGC DL containers have some common info they output at entrypoint - something like this might be a good place to put that

aleksficek:
Closes #96 

Simple instruction added during runtime container execution.

### DIFF
--- a/context/.start_jupyter_run_in_rapids.sh
+++ b/context/.start_jupyter_run_in_rapids.sh
@@ -3,7 +3,7 @@
 conda activate rapids
 /rapids/utils/start_jupyter.sh > /dev/null
 echo "Notebook server successfully started, a JupyterLab instance has been executed!"
-echo "Make local folders visible by mapping to /rapids/notebook"
+echo "Make local folders visible by volume mounting to /rapids/notebook"
 echo "To access visit http://localhost:8888 on your host machine."
 echo 'Ensure the following arguments to "docker run" are added to expose the server ports to your host machine:
    -p 8888:8888 -p 8787:8787 -p 8786:8786'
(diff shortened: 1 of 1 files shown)


context/.start_jupyter_run_in_rapids.sh
```diff
@@ -2,7 +2,8 @@
 . /opt/conda/etc/profile.d/conda.sh
 conda activate rapids
 /rapids/utils/start_jupyter.sh > /dev/null
-echo "Notebook server successfully started!"
+echo "Notebook server successfully started, a JupyterLab instance has been executed!"
+echo "Make local folders visible by mapping to /rapids/notebook"
```
ajschmidt8, line 6:
```suggestion
echo "Make local folders visible by volume mounting to /rapids/notebook"
```

PR #135, opened 2020-08-05
[REVIEW] Add custreamz to devel

dillon-cullinan:


mike-wendt:
Are there any changes that we need to make to `rapids-build-env` to support this? Or is it just the extra build step?

dillon-cullinan:
@mike-wendt build environment should include all dependencies already (only needs `streamz`) so it should be good to go!

mike-wendt:
@dillon-cullinan looks like you need to update the docker files as well

PR #137, opened 2020-08-06
Enable Jupyter to run in foreground with environment variable

### LINKED ISSUE #133: [BUG] Jupyter cannot block and logs are sent to /dev/null
**Describe the bug**

When working on the RAPIDS Helm Chart we have a pod which will be running Jupyter as a daemon. Ideally this pod would simply run Jupyter as the main blocking process.

The current config for the `runtime` image is for Jupyter to be run during the entrypoint with nohup and all stdout/stderr is sent to `/dev/null`.

The default command is `bash` which just exits if the container is not running in interactive mode.

In order for us to block on the Jupyter process we currently have to override the entrypoint and command with a new one which runs Jupyter in the foreground. This results in duplication of the Jupyter startup command in the helm chart. Alternatively we could instead tail the nohup output as the command in order to block, but this isn't possible due to all output going to `/dev/null`.

```
docker run --rm 0.14-cuda10.1-runtime-ubuntu18.04-py3.7
```

Container just exits.

**Expected behavior**

I would expect to be able to run the container with Jupyter as the foreground process.

A few ways this could be achieved:
- Pipe the output to somewhere like `/var/log/jupyter.log` so that we can override the command as `tail -f /var/log/jupyter.log`
- Run Jupyter in the foreground by default without nohup and output redirection
- Split the `start_jupyter.sh` script into a foreground and background version, allowing us to override the entrypoint/command to the foreground version.

**Environment details (please complete the following information):**
Any
 

**Additional context**
NA

cc @quasiben

ajschmidt8:
This PR addresses #133. It enables the `jupyter-lab` server to run in the foreground if the environment variable `JUPYTER_FG` is passed to `docker run`.

i.e: `docker run --gpus all -it -e JUPYTER_FG=true --rm jtest:latest`

image

and without the var: `docker run --gpus all -it --rm jtest:latest`

image

Closes #133.

### DIFF
--- a/context/.start_jupyter_run_in_rapids.sh
+++ b/context/.start_jupyter_run_in_rapids.sh
@@ -1,10 +1,18 @@
 #!/bin/bash
 . /opt/conda/etc/profile.d/conda.sh
 conda activate rapids
-/rapids/utils/start_jupyter.sh > /dev/null
+
+# Run Jupyter in foreground if $JUPYTER_FG is set
+if [[ -v JUPYTER_FG ]]; then
+   OUTPUT="/dev/tty"
+else
+   OUTPUT="/dev/null"
+fi
+
+source /rapids/utils/start_jupyter.sh > "${OUTPUT}"
 echo "Notebook server successfully started, a JupyterLab instance has been executed!"
 echo "Make local folders visible by volume mounting to /rapids/notebook"
 echo "To access visit http://localhost:8888 on your host machine."
 echo 'Ensure the following arguments to "docker run" are added to expose the server ports to your host machine:
    -p 8888:8888 -p 8787:8787 -p 8786:8786'
-exec "$@" 
+exec "$@"
--- a/context/.start_jupyter_run_in_rapids.sh
+++ b/context/.start_jupyter_run_in_rapids.sh
@@ -2,11 +2,11 @@
 . /opt/conda/etc/profile.d/conda.sh
 conda activate rapids
 
+OUTPUT="/dev/null"
+
 # Run Jupyter in foreground if $JUPYTER_FG is set
 if [[ -v JUPYTER_FG ]]; then
    OUTPUT="/dev/tty"
-else
-   OUTPUT="/dev/null"
 fi
 
 source /rapids/utils/start_jupyter.sh > "${OUTPUT}"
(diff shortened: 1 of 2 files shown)


ajschmidt8:
So technically this isn't a bulletproof solution since the `jupyter-lab` server could stop (i.e. crash), and the `tail` command would continue running (I think). But this approach was suggested by @jacobtomlinson who opened the initial issue, so if it meets his requirements than I think it's fine.

An alternative might be to have the `jupyter-lab` server run in the foreground if a particular environment variable is passed.

For example:


```sh
docker run -e "JUPYTER_FG=1" -it --rm rapidsai/rapidsai-nightly:0.15-cuda10.1-runtime-ubuntu18.04-py3.7
```

jacobtomlinson:
Thanks for raising this @ajschmidt8.

I think you're right that my suggestion wasn't ideal. I think a better description of my requirements are as follows:

- The docker log file should show the Jupyter log output
- The container should block on Jupyter Lab so that it can be run as a daemon
- Signals should be passed to Jupyter Lab (not `tail` as I initially suggested)
- The container lifecycle should depend on Jupyter Lab's lifecycle

All of these would be met if Jupyter was run as the foreground process.

Two solutions that jump to mind would be:

1. Use an environment variable to toggle foreground/background running as @ajschmidt8 suggests
2. Add another image named `{rapids-version}-{cuda-version}-jupyter-{distro-version}-{python-version}` which runs Jupyter in the foreground

In Dask we have gone for option 2 and use [daskdev/dask-notebook](https://hub.docker.com/r/daskdev/dask-notebook) for the Jupyter pod in the Helm Chart. In the RAPIDS helm chart (which builds on the Dask one) we use the `runtime` image for the scheduler, workers and jupyter pods.

context/.start_jupyter_run_in_rapids.sh
```diff
@@ -1,10 +1,18 @@
 #!/bin/bash
 . /opt/conda/etc/profile.d/conda.sh
 conda activate rapids
-/rapids/utils/start_jupyter.sh > /dev/null
+
+OUTPUT="/dev/null"
+
+# Run Jupyter in foreground if $JUPYTER_FG is set
+if [[ -v JUPYTER_FG ]]; then
```
raydouglass, line 8: Any reason specifically to check if its set versus set to something like `true`? I just don't want someone to do `JUPYTER_FG=false` when that would still trigger the behavior.

ajschmidt8: good call, I'll push an update to specifically check for `true`.

ajschmidt8: updated:

image

PR #141, opened 2020-08-11
Removed some if statement duplication and pipe to tty

jacobtomlinson:
Made a few changes which should fix the problem I've found in #133 and also condensed the logic a little.

- Instead of piping to `/dev/tty` I've changed this to not pipe at all.
- Simplified `.start_jupyter_run_in_rapids.sh ` and moved all the foreground/background logic to `start_jupyter.sh`.
- Moved background logic into else statement. I feel this makes things more readable as it's easy to miss the short circuit `exit 0` line.
- Remove the command printing and used `set -x` instead. I've just done this for the one line being printed but it may be helpful to just do this for the whole entrypoint script.

### DIFF
--- a/context/start_jupyter.sh
+++ b/context/start_jupyter.sh
@@ -5,9 +5,7 @@ if [[ "${JUPYTER_FG}" == "true" ]]; then
    jupyter-lab --allow-root --ip=0.0.0.0 --no-browser --NotebookApp.token=''
    exit 0
 else
-   set -x
    nohup jupyter-lab --allow-root --ip=0.0.0.0 --no-browser --NotebookApp.token='' > /dev/null 2>&1 &
-   set +x
 
    echo "Notebook server successfully started, a JupyterLab instance has been executed!"
    echo "Make local folders visible by volume mounting to /rapids/notebook"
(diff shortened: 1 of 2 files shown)


context/start_jupyter.sh
```diff
@@ -4,9 +4,15 @@
 if [[ "${JUPYTER_FG}" == "true" ]]; then
    jupyter-lab --allow-root --ip=0.0.0.0 --no-browser --NotebookApp.token=''
    exit 0
+else
+   set -x
+   nohup jupyter-lab --allow-root --ip=0.0.0.0 --no-browser --NotebookApp.token='' > /dev/null 2>&1 &
+   set +x
```
ajschmidt8, line 10:
```suggestion
   nohup jupyter-lab --allow-root --ip=0.0.0.0 --no-browser --NotebookApp.token='' > /dev/null 2>&1 &
```

might be better to just remove the `set -/+ x` here. looks like it clutters up the output message a bit. I'm not sure why it was ever included to be honest.

image

jacobtomlinson: Fine with me!

PR #142, opened 2020-08-14
[BUG] Run the dataset download script for cugraph prior to running NB tests

rlratzel:
This is needed since newer cugraph NBs require extra datasets.  The script is safe to run multiple times since it only downloads if the files aren't already present.

This is part of a fix for several "file not found" errors from cugraph NBs, another change may still be needed in the NBs themselves.


rlratzel:
NOTE: NOTE: [This PR](https://github.com/rapidsai/cugraph/pull/1083) also needs to be merged in order to support the nightly NB test runs for cugraph.

PR #143, opened 2020-08-14
[REVIEW] Update Image Entrypoints

ajschmidt8:
This PR updates the entrypoint in our images. The new entrypoint contains some instructions to create a new user with an id that can (optionally) be passed into the container at runtime (i.e. `docker run -e HOST_USER_ID=$(id -u $USER)...`). This ensures that volume files created from within the container are writeable/deleteable from users outside of the container.

The new entrypoint scripts replace `.run_in_rapids.sh`, `start_jupyter.sh`, and `.start_jupyter_run_in_rapids.sh` so those files have therefore been deleted.

Additionally, `sudo` has been installed on all `base`, `runtime`, and `devel` images.

Finally, the stable DockerHub readmes have been updated as part of this PR.


ajschmidt8:
still need to:

- [ ] update docs
- [ ] validate runtime/devel images locally

mike-wendt:
@ajschmidt8 good work, let's target for 0.16 and use this PR for testing as we get the same PR builds as in gpuci-build-environment

context/run_commands.sh
```diff
@@ -0,0 +1,18 @@
...
+set -e
+
+# give sudo permission for rapids user to run apt/yum (user creation is postponed
+# to the entrypoint, so we can create a user with the same id as the host)
+if cat /etc/os-release | grep centos > /dev/null; then
+  PKG_MGR="yum";
+else
+  PKG_MGR="apt";
+fi
+echo "rapids ALL=NOPASSWD: /usr/bin/${PKG_MGR}" >> /etc/sudoers
```
raydouglass, line 11: Any reason why its limited to just the package manager instead of `ALL`?

ajschmidt8: nope, no reason in particular. this was copied from the set of files that Mike asked me to reference, but I'm fine with changing it if you think it needs to be adjusted

https://github.com/mike-wendt/docker-images/blob/master/scripts/run_commands#L31

PR #145, opened 2020-08-17
FIX Paths and Symlinks for Centos7

aleksficek:
Docker cuDF tests are currently failing on centos. This is because the tests use the system gcc versions which are out of date and instead of the manually installed gcc7.

This PR fixes ccache to symlink to the correct gcc version on centos.


aleksficek:
Fix should be done in integration repo

PR #153, opened 2020-08-26
[FIX] Use build-arg for RAPIDS_VER in Dockerfiles

ajschmidt8:
This PR ensures that the `RAPIDS_VER` variable value uses Docker build arguments instead of Jinja variables. This allows different versions of the containers to be built on specific version branches (i.e. `0.16` nightlies can be successfully built on `branch-0.15`).

The changes include:

- Renaming `RAPIDS_VERSION` to `DEFAULT_ RAPIDS_VERSION` in `settings.yaml`
- Updating the `ARG RAPIDS_VER=0.15*` Dockerfile step to just `ARG RAPIDS_VER` so that it uses the `RAPIDS_VER` value that occurs before the Dockerfile's `FROM` statement, which can be passed in via a build-argument
- Make sure any conda operations include the `*` since it was removed from the variable in the previous step. Use quotes to prevent shell expansion (i.e. `rapids=${RAPIDS_VER}` => `"rapids=${RAPIDS_VER}*"`
- Replace all Jinja `{{ RAPIDS_VERSION }}` instances with the Docker argument, `${RAPIDS_VER}`


generated-dockerfiles/centos7-base.Dockerfile
```diff
@@ -14,7 +14,7 @@ ARG FROM_IMAGE=gpuci/rapidsai
 FROM ${FROM_IMAGE}:${RAPIDS_VER}-cuda${CUDA_VER}-base-${LINUX_VER}-py${PYTHON_VER}
 
 ARG DASK_XGBOOST_VER=0.2*
-ARG RAPIDS_VER=0.15*
+ARG RAPIDS_VER
```
mike-wendt, line 17: Shouldn't this have a value? That way if you do just `docker build` without args it will build with all of the default values?

ajschmidt8: > Shouldn't this have a value? That way if you do just `docker build` without args it will build with all of the default values?

It will use the value from before the `FROM` statement [here](https://github.com/rapidsai/docker/blob/6a44f90ecdc561f1b67cfff21e4ae35c5108e2d6/generated-dockerfiles/centos7-devel.Dockerfile#L13). 
(src: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact).

I was able to confirm this locally with the similar example below:

```dockerfile
ARG ALP_VER=3

FROM alpine:${ALP_VER}

ARG ALP_VER

RUN echo "${ALP_VER}"

```

I ran both `docker build .` and `docker build --build-arg "ALP_VER=3.12" .` and got the expected values.

mike-wendt: Got it my mistake I see you're defining this to use it again later

PR #166, opened 2020-09-11
Fix cublas include error

ajschmidt8:
As seen in the screenshot below, `nvidia/cuda:10.1-devel-centos7` images have the `cublas_v2.h` files in the `/usr/local/cuda-10.2/include` folder instead of `/usr/local/cuda-10.1/include`. This causes "missing cublas_v2.h" errors in our `devel` build Jenkins jobs [as shown here](https://gpuci.gpuopenanalytics.com/job/rapidsai/job/docker/job/rapidsai-devel/120/).

After reporting this to Jesus Alvarez, maintainer of the `nvidia/cuda` images, he mentioned that this was intentional and pointed me to the [explanation shown here](https://gitlab.com/nvidia/container-images/cuda/-/issues/88#note_411485001).

As a result, our images need to be patched with the changes in this PR to get the `centos-devel` builds working again.

image

These changes were tested locally and confirmed working.


ajschmidt8:
moving PR to `gpuci-build-environment` repo to fix this in parent images

PR #170, opened 2020-09-17
Enable packages to be installed via runtime environment variables

### LINKED ISSUE #25: [FEA] Add extra packages in docker entrypoint
I would like to use the rapids docker image more interchangeable with the [dask docker image](https://github.com/dask/dask-docker). 

**Describe the solution you'd like**
The dask image (along with others) implements an [entrypoint script](https://github.com/dask/dask-docker/blob/master/base/prepare.sh) which installs extra packages from environment variables (`EXTRA_APT_PACKAGES`, `EXTRA_CONDA_PACKAGES`, `EXTRA_PIP_PACKAGES`). This allows users to customize their containers without having to build their own.

**Describe alternatives you've considered**
The alternative to this is for people to build their own docker images backed on the rapids image and host it somewhere like Docker Hub. It would be good if users could get started with the docker image without prior knowledge of docker.

ajschmidt8:
Closes #25.

This PR enables packages to be installed via environment variables that are assigned during `docker run`.

For example:

```sh
docker run -e EXTRA_YUM_PACKAGES="nano vim"...
```

In particular, the following runtime environment variables were introduced:

- `EXTRA_APT_PACKAGES`
- `EXTRA_YUM_PACKAGES`
- `EXTRA_CONDA_PACKAGES`
- `EXTRA_PIP_PACKAGES`

Additionally, a YAML file of dependencies can be mounted to `/opt/rapids/environment.yml` and the dependencies will be installed.


context/packages.sh
```diff
@@ -0,0 +1,38 @@
...
+fi
+
+if [ -e "/opt/rapids/environment.yml" ]; then
+    echo "environment.yml found. Installing packages"
+    conda env update -f /opt/rapids/environment.yml
+fi
+
+if [ "$EXTRA_CONDA_PACKAGES" ]; then
+    echo "EXTRA_CONDA_PACKAGES environment variable found. Installing packages."
+    conda install -y $EXTRA_CONDA_PACKAGES
```
jacobtomlinson, line 32: Just checking is the rapids environment already activated here?

ajschmidt8: yup! that's correct. it gets run after a `conda activate rapids` call [here](https://github.com/ajschmidt8/docker/blob/user-pkgs/context/entrypoint.sh#L17-L20)

raydouglass (approved):

mike-wendt (approved):

jacobtomlinson (approved):

PR #175, opened 2020-10-06
Update to branch-0.17

ajschmidt8:
This PR makes the necessary changes to `branch-0.17` to update it to use the `0.17` library versions.

It also removes the `RAPIDS_CHANNEL` axis value since it seems to be extraneous. `RAPIDS_CHANNEL` was used in `ci/gpuci/run.sh` to add a `--build-arg RAPIDS_CHANNEL=${RAPIDS_CHANNEL}` argument to Docker. However, our Docker images do not use the `RAPIDS_CHANNEL` build argument. I assume this was an oversight during a copy-paste from `gpuci-build-environment` ([src](https://github.com/rapidsai/gpuci-build-environment/blob/branch-0.16/ci/gpuci/run.sh#L75-L81)).


ci/axis/base-runtime.yaml
```diff
@@ -11,9 +11,7 @@ IMAGE_TYPE:
 
 RAPIDS_VER:
   - 0.16
-
-RAPIDS_CHANNEL:
```
mike-wendt, line 15: We use `RAPIDS_CHANNEL` when we do a release, as we need to switch to `rapidsai` from `rapidsai-nightly`

mike-wendt (approved):

mike-wendt:
> It also removes the `RAPIDS_CHANNEL` axis value since it seems to be extraneous. `RAPIDS_CHANNEL` was used in `ci/gpuci/run.sh` to add a `--build-arg RAPIDS_CHANNEL=${RAPIDS_CHANNEL}` argument to Docker. However, our Docker images do not use the `RAPIDS_CHANNEL` build argument. I assume this was an oversight during a copy-paste from `gpuci-build-environment` ([src](https://github.com/rapidsai/gpuci-build-environment/blob/branch-0.16/ci/gpuci/run.sh#L75-L81)).

I think it's an artifact of when we used to also define the `.condarc` files in this repo. Now that they have been moved to `gpuci-build-environment` we do not need the definitions here as they are baked into the `FROM` images

PR #225, opened 2020-12-07
add docker hub readme autogeneration action

jolorunyomi:
# Summary
This PR adds DockerHub readme autogeneration via a GH Action.

# Testing Evidence
https://github.com/rapidsai/docker/pull/225/checks?check_run_id=1512216313

### DIFF
--- a/.github/workflows/dockerhub-readme.yml
+++ b/.github/workflows/dockerhub-readme.yml
@@ -1,0 +1,64 @@
+name: Generate DockerHub READMEs
+
+on:
+  push:
+    branches:
+      - branch-*
+    paths:
+      - dockerhub-readme/
+      - .github/workflows/dockerhub-readme.yml
+
+jobs:
+  readme:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v2
+        
+      - name: Docker Hub Description >>> Blazing dev nightly
+        uses: peter-evans/dockerhub-description@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+          repository:  rapidsai/blazingsql-dev-nightly
+          readme-filepath: dockerhub-readme/Dockerhub_blazing-dev-nightly.md
+          
+      - name: Docker Hub Description >>> Blazing dev
+        uses: peter-evans/dockerhub-description@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+          repository:  rapidsai/blazingsql-dev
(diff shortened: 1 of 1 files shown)


.github/workflows/dockerhub-readme.yml
```diff
@@ -0,0 +1,64 @@
+name: Generate DockerHub READMEs
+
+'on':
```
ajschmidt8, line 3:
```suggestion
on:
```

Are the quotes here needed? I've never had to use quotes for the `on` key when doing GitHub Actions ([src1](https://github.com/rapidsai/docker/blob/branch-0.17/.github/workflows/dockerfiles.yaml#L3), [src2](https://github.com/rapidsai/gpuci-scripts/blob/branch-0.17/.github/workflows/jenkins_plugins.yaml#L3)).

.github/workflows/dockerhub-readme.yml
```diff
@@ -0,0 +1,64 @@
...
+  push:
+    branches:
+      - main
```
ajschmidt8, line 6: This is tricky. `main` is used for all of our stable releases. However, the `README.md`s can change in between stable releases (i.e. for the nightlies) and we'd still want this GitHub Action to trigger. Ideally, we'd set this to be our default branch (i.e. `branch-0.17`), but this is problematic since our default branch changes regularly. We should see if there's some way for us to fix this.

.github/workflows/dockerhub-readme.yml
```diff
@@ -0,0 +1,64 @@
...
+    paths:
+      - dockerhub-readme/
```
ajschmidt8, line 8: nice.

ajschmidt8 (changes requested):
nice, looks good. let's see if there's a way to fix the branch issue that I described.

.github/workflows/dockerhub-readme.yml
ajschmidt8: I think we could get away with using `branches: branch-*` for the branch here. The `rapidsai/docker` repo doesn’t follow the same burndown process as the library repos (where `branch-0.17` and `branch-0.18` can be active at the same time). So since this repo only ever has a single active `branch-0.xx` branch, this should probably suffice.

jolorunyomi: we could listen to every push, check default branch with git and abort on that. it will be spammy though.

.github/workflows/dockerhub-readme.yml
jolorunyomi: Yeah I found that depending on how the yaml is deserialized, `on` may be converted to `true`.

E.g 
image

But I have unquoted it though.

.github/workflows/dockerhub-readme.yml
jolorunyomi: updated with `branches-*`

ajschmidt8 (approved):

PR #226, opened 2020-12-09
Add Dask Labextension config

jacobtomlinson:
Add some config for the Dask Jupyter Lab Extension to use `dask_cuda.LocalCUDACluster` by default.

The result of this means clicking the "New" button in the Dask extension will create a `LocalCUDACluster` instead of a `LocalCluster`.

image

cc @pentschev


pentschev (approved):
This is a great idea, thanks @jacobtomlinson ! I'm not sure if that's possible without changes in Dask itself, but could we also change the text to "Start New Dask _CUDA_ Cluster"? This is only a cosmetic clarification change, and I'm not sure if that's important, certainly I'd ignore it if that requires changes elsewhere.

jacobtomlinson:
It's not possible right now. It's a good idea though. 

I'd like to improve the cluster creation in the Lab Extension generally at some point. So maybe it's something to tackle then.

ajschmidt8:
@jacobtomlinson, sorry for the delay here. can you pull the latest changes into your branch here? I updated the GitHub Action name recently. I can merge after that.

jacobtomlinson:
Thanks @ajschmidt8. Merged, regenerated and pushed.

ajschmidt8 (approved):

PR #250, opened 2021-01-24
Fix #249: Display volume binding message only if not done by the user

### LINKED ISSUE #249: [FEA] Display volume binding message only if not done by the user
**Is your feature request related to a problem? Please describe.**
Currently, the following message is always displayed regardless if the user has mounted a volume on the desired path:

> Make local folders visible by bind mounting to /rapids/notebooks/host

**Describe the solution you'd like**
It would be great to double-check if the path exists before displaying the message.

miguelusque:
Hi!

I have added a small check before displaying the following message, to avoid displaying it if the user has already bound a volume on the right path.

> Make local folders visible by bind mounting to /rapids/notebooks/host

Hope it helps!
Miguel


ajschmidt8 (approved):
Awesome! Thanks!

PR #252, opened 2021-01-27
Install Jupyter kernel discovery and use nbgitpuller for examples

jacobtomlinson:
I'm working on using the RAPIDS Docker images with Jupyter Hub. 

Two small tools that would be really helpful are `nb_conda_kernels` for conda environment discovery in Jupyter and `nbgitpuller` for downloading example notebooks.

We already download the example notebooks in the RAPIDS runtime image. But `nbgitpuller` is a small utility made by the Jupyter folks to automate this, and specifically automate updating these examples. So when we launch the docker image with JupyterHub we can easily update the examples at runtime.

### DIFF
--- a/templates/rapidsai-core/partials/install_notebooks.dockerfile.j2
+++ b/templates/rapidsai-core/partials/install_notebooks.dockerfile.j2
@@ -7,7 +7,7 @@ RUN gpuci_conda_retry install -y -n rapids \
         "rapids-notebook-env=${RAPIDS_VER}*"
 
 {# Install jupyter lab stuff #}
-RUN gpuci_conda_retry install -y -n rapids jupyterlab-nvdashboard
+RUN gpuci_conda_retry install -y -n rapids jupyterlab-nvdashboard nb_conda_kernels nbgitpuller
 
 RUN source activate rapids \
   && jupyter labextension install @jupyter-widgets/jupyterlab-manager dask-labextension jupyterlab-nvdashboard
@@ -19,9 +19,7 @@ ENV DASK_LABEXTENSION__FACTORY__CLASS="LocalCUDACluster"
 {# Install notebooks repo #}
 RUN cd ${RAPIDS_DIR} \
   && source activate rapids \
-  && git clone -b ${BUILD_BRANCH} --depth 1 --single-branch https://github.com/rapidsai/notebooks.git \
-  && cd notebooks \
-  && git submodule update --init --remote --no-single-branch --depth 1
+  && gitpuller https://github.com/rapidsai/notebooks ${BUILD_BRANCH} notebooks
 
 {# Add test file for testing notebooks from within the container #}
 COPY test.sh /
--- a/generated-dockerfiles/rapidsai-core_centos7-devel.Dockerfile
+++ b/generated-dockerfiles/rapidsai-core_centos7-devel.Dockerfile
@@ -77,7 +77,7 @@ RUN gpuci_conda_retry install -y -n rapids \
     && gpuci_conda_retry remove -y -n rapids --force-remove \
         "rapids-notebook-env=${RAPIDS_VER}*"
 
-RUN gpuci_conda_retry install -y -n rapids jupyterlab-nvdashboard
+RUN gpuci_conda_retry install -y -n rapids jupyterlab-nvdashboard nb_conda_kernels nbgitpuller
 
 RUN source activate rapids \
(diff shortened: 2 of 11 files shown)


quasiben (approved):
+1 to using `nbgitpuller` I think this also what `nb_conda_kernels` use as well

templates/rapidsai-core/partials/install_notebooks.dockerfile.j2
```diff
@@ -7,7 +7,7 @@ RUN gpuci_conda_retry install -y -n rapids \
         "rapids-notebook-env=${RAPIDS_VER}*"
 
 {# Install jupyter lab stuff #}
-RUN gpuci_conda_retry install -y -n rapids jupyterlab-nvdashboard
+RUN gpuci_conda_retry install -y -n rapids jupyterlab-nvdashboard nb_conda_kernels nbgitpuller
```
ajschmidt8, line 10: these seem like cool additions, but i have some questions/comments

1) The addition of `nb_conda_kernels` should be fine. However, you've uncovered a bit of tech debt here on our end. We try to keep all the notebook dependencies in the [rapids-notebook-env meta-package in the integration repo](https://github.com/rapidsai/integration/blob/branch-0.18/conda/recipes/rapids-notebook-env/meta.yaml). So ideally `jupyterlab-nvdashboard` and `nb_conda_kernels` would both be added there.

2) Our [notebooks repo](https://github.com/rapidsai/notebooks) consist mostly of a bunch of git-submodules. When I tested the new `gitpuller` command locally, it didn't seem to also clone in the sub-modules (see screenshot below). After exploring the usage section in `gitpuller --help`, I didn't see anything that mentioned submodules so I'm not sure if it's compatible. Also, the `notebooks` repo (like other RAPIDS repos) has a default branch that changes every release, so I'm not sure how this would affect the functionality of `gitpuller`. Any thoughts?

image

ajschmidt8 (changes requested):
couple questsions/comments below

templates/rapidsai-core/partials/install_notebooks.dockerfile.j2
jacobtomlinson: 1. Ah ok that's good to know. Should I open a PR over there?
2. Interesting that it doesn't grab the submodules. Looks like there is an open PR to resolve jupyterhub/nbgitpuller#123. Also my intention on using the `${BUILD_BRANCH}` env var was that the correct branch would be pulled.

ajschmidt8: > 1. Ah ok that's good to know. Should I open a PR over there?
> 2. Interesting that it doesn't grab the submodules. Looks like there is an open PR to resolve [jupyterhub/nbgitpuller#123](https://github.com/jupyterhub/nbgitpuller/pull/123). Also my intention on using the `${BUILD_BRANCH}` env var was that the correct branch would be pulled.

1. Yes, if you don't mind that would be great. You could then also remove the entire `RUN gpuci_conda_retry install` line from this PR.
2. Whoops. For this comment, I was referring to this statement in the PR body: `So when we launch the docker image with JupyterHub we can easily update the examples at runtime`. I was thinking about the following for both nightly and stable images:
- **Nightlies** - nightlies are rebuilt every night, so therefore the notebooks are usually up to date when you pull the latest nightly
- **Stable images** - whenever we publish a stable release with new images, the active development branch changes. So if `BUILD_BRANCH` is equal to `branch-0.19` for instance, then after the `0.19` release, all new development will occur on the `branch-0.20` branch. I'm not sure how/if `nbgitpuller` will handle this branch change. I assume (though I didn't confirm this in any way) that `nbgitpuller` will only update the notebooks according to the branch in which it was originally invoked with (i.e. `branch-0.19` in the previous example). If this is the case, then I think that `nbgitpuller` would only be useful in the instance where users download an early nightly container and want to update their notebooks to the latest notebooks _for the RAPIDS version of their current container_ (and not necessarily the latest _overall_ notebooks, which could potentially be on a new branch). Hopefully that makes sense. `nbgitpuller` is small enough that it's probably fine to include in our images regardless, but all of this is moot until they get that submodules PR merged :slightly_smiling_face: .

ajschmidt8:
@jacobtomlinson, I'm not sure if this PR is still relevant with the recent overhaul of the Docker images.

Can we close this?

If these features will still work with the new images, can you adapt your PR to the changes that were recently merged?

jacobtomlinson:
I think these changes would still be useful, but the exact use case that was motivating this in 2021 is gone now so I'll probably follow up if/when it comes up again.

PR #257, opened 2021-02-04
Add verbose flag on build.sh calls

mike-wendt:
This is to help with debugging current errors after conda gcc 9.3.0 build stack was merged, but we are still building with system level gcc 7.5.0


mike-wendt:
Debugging complete, closing

PR #260, opened 2021-02-11
[REVIEW] FIX Reproduce symlinks to ccache before source builds

dillon-cullinan:
Fixes the problems with builds losing access to compilers. It seems like installing new packages activates the `deactivate.sh` script from `rapids-scout-local` which removes symlinks to ccache which source builds rely on for finding compilers.

Tested locally with a full docker build, succeeded. Changes are OS/CUDA/Python agnostic so it should work across all of our supported software.

### DIFF
--- a/generated-dockerfiles/rapidsai-clx_centos7-devel.Dockerfile
+++ b/generated-dockerfiles/rapidsai-clx_centos7-devel.Dockerfile
@@ -34,6 +34,10 @@ RUN source activate rapids && \
     pip install mockito && \
     pip install wget
 
+RUN ln -s "$(which ccache)" "/usr/local/bin/gcc" \
+    && ln -s "$(which ccache)" "/usr/local/bin/g++" \
+    && ln -s "$(which ccache)" "/usr/local/bin/nvcc"
+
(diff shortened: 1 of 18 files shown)


ajschmidt8 (changes requested):
I'm not sure this PR addresses the root cause of the missing symlinks. I'm also experiencing disappearing symlinks in the `core` images (see screenshot below). It seems that the symlinks disappear when packages are installed via `conda` (but only certain package installations, oddly). This PR only replaces the symlinks, but doesn't seem to prevent them from being removed again after another `conda` install command. I think we need to re-evaluate `scout` and the [deactivate.sh](https://github.com/rapidsai/scout/blob/branch-0.18/conda/recipe/scout-local/deactivate.sh#L9-L11) script to determine if that's causing the issue here.

image

ajschmidt8:
also seeing other `gcc` related errors in the conda `base` environment

image

dillon-cullinan:
Closing this as the changes made in `rapids-scout-local` package should resolve the build problem: https://github.com/rapidsai/scout/pull/4

PR #285, opened 2021-04-01
[REVIEW] FIX Handle centos for CVE patching

dillon-cullinan:
Added logic to use yum for package patching in centos.

### DIFF
--- a/templates/rapidsai-core/Base.dockerfile.j2
+++ b/templates/rapidsai-core/Base.dockerfile.j2
@@ -42,12 +42,6 @@ COPY packages.sh /opt/docker/bin/
 {# Cleaup conda and set ACLs for all users #}
 {% include 'partials/cleanup-chmod.dockerfile.j2' %}
 
-{% if "centos" in os %}
-RUN yum -y update && yum -y upgrade
-{% else %}
-RUN apt -y update && apt -y upgrade
-{% endif %}
-
 WORKDIR ${RAPIDS_DIR}
 
 COPY entrypoint.sh /opt/docker/bin/entrypoint
--- a/templates/rapidsai-core/Base.dockerfile.j2
+++ b/templates/rapidsai-core/Base.dockerfile.j2
@@ -42,7 +42,11 @@ COPY packages.sh /opt/docker/bin/
 {# Cleaup conda and set ACLs for all users #}
 {% include 'partials/cleanup-chmod.dockerfile.j2' %}
 
-RUN apt update && apt -y upgrade
+{% if "centos" in os %}
+RUN yum -y update && yum -y upgrade
+{% else %}
+RUN apt -y update && apt -y upgrade
+{% endif %}
 
 WORKDIR ${RAPIDS_DIR}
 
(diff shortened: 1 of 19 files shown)


templates/rapidsai-core/Base.dockerfile.j2
```diff
@@ -42,7 +42,11 @@ COPY packages.sh /opt/docker/bin/
 {# Cleaup conda and set ACLs for all users #}
 {% include 'partials/cleanup-chmod.dockerfile.j2' %}
 
-RUN apt-get update && apt-get -y upgrade
+{% if "centos" in os %}
+RUN yum -y update && yum -y upgrade
+{% else %}
+RUN apt -y update && apt -y upgrade
+{% endif %}
```
ajschmidt8, line 49: Looks like this same code is duplicated across the `Base`, `Runtime`, and `Devel` templates, so it'd likely be a good candidate for a Jinja partial. Can you add it to the `partials/patch.dockerfile.j2` file? This is already included in all 3 image types and was made particularly for patching CVEs.

ajschmidt8 (changes requested):

templates/rapidsai-core/partials/patch.dockerfile.j2
```diff
@@ -3,3 +3,9 @@
 {# Patch for CVE-2020-8116 https://github.com/advisories/GHSA-ff7x-qrg7-qggm #}
 RUN source activate rapids \
     && npm i -g npm@">=7"
+
+{% if "centos" in os %}
+RUN yum -y update && yum -y upgrade
+{% else %}
+RUN apt-get -y update && apt-get -y upgrade
```
ajschmidt8, line 10:
```suggestion
RUN yum -y upgrade
{% else %}
RUN apt-get update && apt-get -y upgrade
```

Two changes here:

1) Removes `-y` from `apt-get update` since it's not necessary
2) Removes `yum -y update`. From the sources linked below (and my local testing) it seems that `yum update` and `yum upgrade` do nearly the same thing except that `yum upgrade` will remove any obsolete packages when it's done running (whereas `yum update` does not remove obsolete packages). ([src1](https://unix.stackexchange.com/a/55778), [src2](https://embeddedinventor.com/yum-update-vs-yum-upgrade-differences-explained/#:~:text=What%20is%20the%20difference%20between,obsolete%20packages%20from%20the%20system.))

ajschmidt8 (changes requested):

templates/rapidsai-core/partials/patch.dockerfile.j2
```diff
@@ -3,3 +3,9 @@
 {# Patch for CVE-2020-8116 https://github.com/advisories/GHSA-ff7x-qrg7-qggm #}
 RUN source activate rapids \
     && npm i -g npm@">=7"
+
+{% if "centos" in os %}
+RUN yum -y upgrade
+{% else %}
+RUN apt-get update && apt-get -y upgrade
```
ajschmidt8, line 10:
```suggestion
RUN yum -y upgrade \
    && yum clean all
{% else %}
RUN apt-get update \
    && apt-get -y upgrade \
    && rm -rf /var/lib/apt/lists/*
```

Cleans up `apt` and `yum` cache to reduce image size.

src: 
- https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get
- https://hub.docker.com/_/centos

ajschmidt8 (changes requested):

ajschmidt8 (approved):

PR #307, opened 2021-04-20
Dockerfile ARG fix

ajschmidt8:
This PR updates the Dockerfiles to add some `ARG` statements after the `FROM` statement. This is so that the values of the `ARG`s that are defined before the `FROM` statement also get defined after the `FROM` statement.

Relevant Docker docs describing this behavior: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact


dillon-cullinan (approved):

raydouglass (approved):

ajschmidt8:
Because of the problem fixed in this PR, the following code block likely never worked and therefore could alter how `xgboost` is built moving forward. @raydouglass, any thoughts on this?

https://github.com/rapidsai/docker/blob/branch-0.19/generated-dockerfiles/rapidsai-core_ubuntu18.04-devel.Dockerfile#L175-L199

raydouglass:
> any thoughts on this?

I think it's probably fine. The only difference is `-DCMAKE_CXX_STANDARD:STRING="14"` so I'm not too concerned.

PR #317, opened 2021-04-30
Add `openmpi`

raydouglass:
Adds OpenMPI to base & runtime images to allow for multi-node NGC.

### DIFF
--- a/templates/rapidsai-core/partials/os_pkgs.dockerfile.j2
+++ b/templates/rapidsai-core/partials/os_pkgs.dockerfile.j2
@@ -1,11 +1,13 @@
 {% if "centos" in os %}
 RUN yum install -y \
-      openssh-clients
+      openssh-clients \
+      openmpi-devel
 {% endif %}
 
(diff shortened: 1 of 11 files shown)


Ethyling (approved):

dillon-cullinan (approved):

templates/rapidsai-core/partials/os_pkgs.dockerfile.j2
```diff
@@ -1,11 +1,13 @@
 {% if "centos" in os %}
 RUN yum install -y \
-      openssh-clients
+      openssh-clients \
+      openmpi-devel
```
ajschmidt8, line 4:
```suggestion
      openmpi-devel \
      && yum clean all
```

Looks like we could also add a `yum clean all` here, similar to https://github.com/rapidsai/docker/blob/branch-0.20/templates/rapidsai-core/partials/patch.dockerfile.j2#L9

ajschmidt8 (approved):

PR #323, opened 2021-05-12
ENH Remove ubuntu16.04 images

Ethyling:
Remove ubuntu16.04 images


dockerhub-readme/templates/base.md.j2
```diff
@@ -97,7 +97,7 @@ Many users do not need a specific platform combination but would like to ensure
 {% if is_ngc %}
 - Ubuntu 18.04/20.04 or CentOS 7/8
 {% else %}
-- Ubuntu 16.04/18.04/20.04 or CentOS 7/8
+- Ubuntu 18.04/20.04 or CentOS 7/8
 {% endif %}
```
ajschmidt8, line 101: This conditional looks like it's no longer needed since `ngc` and non-`ngc` now both have the same text.

raydouglass (approved):

PR #333, opened 2021-06-08
Allow Jupyter Lab connections from any origin

jacobtomlinson:
This is required to allow Jupyter to be accessible from behind a reverse proxy with a URL. Otherwise only localhost and IP address connections are accepted.


ajschmidt8 (approved):

jacobtomlinson:
Thanks @ajschmidt8 !

PR #334, opened 2021-06-09
[REVIEW] Ensure no packages from conda defaults channel

dillon-cullinan:
This PR ensures no docker image build succeeds if it contains a package from the conda defaults channel.

Also adds some support in the `generate_dockerfiles.py` file to allow for global partials.

Logic has been tested and confirmed working on runtime containers.

### DIFF
--- a/generated-dockerfiles/rapidsai-clx_centos7-base.Dockerfile
+++ b/generated-dockerfiles/rapidsai-clx_centos7-base.Dockerfile
@@ -20,13 +20,14 @@ RUN gpuci_conda_retry install -y -n rapids -c pytorch \
     "clx=${RAPIDS_VER}" \
     "cudatoolkit=${CUDA_VER}"
 
-
-ENV CONDA_DEFAULT_PKG_NUM=`conda list | awk '{ print $4}' | grep defaults | wc -l`
-RUN if [[ ${CONDA_DEFAULT_PKG_NUM} -eq 0 ]]; then echo "ERROR: Packages from the default conda channel detected"; exit 1; fi
 WORKDIR ${RAPIDS_DIR}
 
 RUN conda clean -afy
 
+
+RUN source activate rapids \
+    && CONDA_DEFAULT_PKG_NUM=`conda list | awk '{ print $4 }' | grep defaults | wc -l` \
+    && if [[ ${CONDA_DEFAULT_PKG_NUM} -ne 0 ]]; then echo "ERROR: Packages from the default conda channel detected"; exit 1; fi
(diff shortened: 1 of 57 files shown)


ajschmidt8:
This test is useful, but I'm not sure it belongs in the Dockerfile steps. I think it'd be more suitable to add to the tests that we run on the images once they're already built. Adding tests to a Dockerfile doesn't seem like a good practice to initiate.

templates/partials/ensure-no-defaults.dockerfile.j2
```diff
@@ -0,0 +1,6 @@
+{# This partial file is used to ensure there are no packages from the defaults conda channel #}
+
+{# Check for defaults packages in rapids env #}
+RUN source activate rapids \
```
raydouglass, line 4: Should we output `conda list` before the check so if it fails we can see what packages were in `defaults`?

raydouglass:
> Adding tests via Dockerfile steps doesn't seem like a good practice to initiate.

I think I agree with this.

generated-dockerfiles/rapidsai-clx_centos7-base.Dockerfile
```diff
@@ -24,6 +24,10 @@ WORKDIR ${RAPIDS_DIR}
 
 RUN conda clean -afy
 
+
+RUN source activate rapids \
+    && CONDA_DEFAULT_PKG_NUM=`conda list | awk '{ print $4 }' | grep defaults | wc -l` \
```
mattf, line 29: fyi: grep defaults | wc -l -> grep -c defaults

templates/partials/ensure-no-defaults.dockerfile.j2
```diff
@@ -0,0 +1,6 @@
...
+    && CONDA_DEFAULT_PKG_NUM=`conda list | awk '{ print $4 }' | grep defaults | wc -l` \
```
mattf, line 5: fyi: `grep defaults | wc -l` is shorter as `grep -c defaults`

templates/partials/ensure-no-defaults.dockerfile.j2
jakirkham, line 4: Should we also check the `base` environment as well?

dillon-cullinan: Ah, that makes sense.

jakirkham: Right Conda's. Sorry for the confusion

PR #339, opened 2021-06-15
Add option to disable quoting in entrypoint exec

Ethyling:
Add option to disable quoting in entrypoint exec

Usage:
```
docker run <image>:<tag> --unquote-exec "cmd"
docker run -e UNQUOTE_EXEC=true <image>:<tag> "cmd"
```

### DIFF
--- a/context/entrypoint.sh
+++ b/context/entrypoint.sh
@@ -17,5 +17,18 @@ source /opt/docker/bin/packages.sh
 SRC_FILE="/opt/docker/bin/entrypoint_source"
 [ -f "${SRC_FILE}" ] && source "${SRC_FILE}"
 
+# Check if we should quote the exec params
+UNQUOTE=false
+if [ "$1" = "--unquote-exec" ]; then
+  UNQUOTE=true
+  shift
+elif [ -n "${UNQUOTE_EXEC}" ] && [[ "${UNQUOTE_EXEC}" =~ ^(true|yes|y)$ ]]; then
+  UNQUOTE=true
+fi
+
 # Run whatever the user wants.
-exec "$@"
+if [ "${UNQUOTE}" = "true" ]; then
+  exec $@
+else
+  exec "$@"
+fi


ajschmidt8 (commented):
this should work, but looks like you just need to flip the logic here. by default it should be quoted. the flag should make it unquoted. so maybe use `--unquoted-cmd` or `--unquoted`.

context/entrypoint.sh
```diff
@@ -17,5 +17,18 @@ source /opt/docker/bin/packages.sh
 SRC_FILE="/opt/docker/bin/entrypoint_source"
 [ -f "${SRC_FILE}" ] && source "${SRC_FILE}"
 
+# Check if we should quote the exec params
+UNQUOTE=false
+if [ "$1" = "--unquote-exec" ]; then
```
jacobtomlinson, line 22: This change is to support where `$@` is already a quoted string. So I'm not sure whether `$1` will actually be `"--unquote-exec"`. It might be `"\"--unquote-exec"`, but I'm honestly not sure.

jacobtomlinson (commented):
Thanks for pushing on this @Ethyling.

I had a go at implementing this myself and I think it may be a little more nuanced. The fact that `$@` can even be quoted is an unusual case, so I think things may not quite work as expected here.

context/entrypoint.sh
Ethyling: I pushed a new version that should work in that case. WDYT?

jacobtomlinson: Thansk @Ethyling. That makes sense, but I also pinged @sriramch who seems happy that the original fix should also work.

Ethyling: Oh ok, so I think I will revert to the previous version then. Thank you!

Ethyling: Done

Ethyling:
@jacobtomlinson Can you double check this? Then we should be good to merge it

jacobtomlinson (approved):

ajschmidt8 (approved):

PR #343, opened 2021-06-22
[REVIEW] DOC Add documentation snippet about rapids environment

dillon-cullinan:
Adds some documentation to help people that want to build images on top of the RAPIDS images.

People have been vocal about facing issues working with RAPIDS images and finding it difficult to properly run their commands in the right environment. Due to its generality, it is also hard to find a solution in the public domain, so directly addressing it in our README should make it easier for users who are interested in doing this.

This README is linked on our homepage under `Getting Started`, so this won't be added specifically to the site.

### DIFF
--- a/dockerhub-readme/templates/base.md.j2
+++ b/dockerhub-readme/templates/base.md.j2
@@ -195,6 +195,10 @@ Notebooks can be found in the following directories within the {{ repo_rapids_ve
 
 For a full description of each notebook, see the [README](https://github.com/rapidsai/notebooks/blob/branch-{{ repo_rapids_version }}/README.md) in the notebooks repository.
 
+### Building on top of RAPIDS images
+
+All RAPIDS images use `conda` as their package manager, and all RAPIDS packages (including source-built) are installed under the `rapids` environment name. If you want to build on top of the RAPIDS images (such as using `FROM`), then it is important to include `source activate rapids` at the start of all `RUN` commands in your `Dockerfile`. Without this, the docker build context will not have access to the RAPIDS libraries, as it uses the `base` environment by default. Examples of this can be found in our own Dockerfiles, which can be found in the [RAPIDS Docker Repository](https://github.com/rapidsai/docker) on GitHub.
+
 ### Custom Data and Advanced Usage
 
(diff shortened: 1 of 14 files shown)


dockerhub-readme/templates/base.md.j2
```diff
@@ -195,6 +195,10 @@ Notebooks can be found in the following directories within the {{ repo_rapids_ve
 
 For a full description of each notebook, see the [README](https://github.com/rapidsai/notebooks/blob/branch-{{ repo_rapids_version }}/README.md) in the notebooks repository.
 
+### Building on top of RAPIDS images
```
ajschmidt8, line 198:
```suggestion
### Extending RAPIDS Images
```

dockerhub-readme/templates/base.md.j2
```diff
@@ -195,6 +195,10 @@ Notebooks can be found in the following directories within the {{ repo_rapids_ve
...
+
+All RAPIDS images use `conda` as their package manager, and all RAPIDS packages (including source-built) are installed under the `rapids` environment name. If you want to build on top of the RAPIDS images (such as using `FROM`), then it is important to include `source activate rapids` at the start of all `RUN` commands in your `Dockerfile`. Without this, the docker build context will not have access to the RAPIDS libraries, as it uses the `base` environment by default. Examples of this can be found in our own Dockerfiles, which can be found in the [RAPIDS Docker Repository](https://github.com/rapidsai/docker) on GitHub.
```
ajschmidt8, line 200:
```suggestion
All RAPIDS images use `conda` as their package manager, and all RAPIDS packages (including source-built) are available in the `rapids` conda environment. If you want to extend RAPIDS images (such as using `FROM`), then it is important to include `source activate rapids` at the start of all `RUN` commands in your `Dockerfile`. Without this, the docker build context will not have access to the RAPIDS libraries, as it uses the `base` environment by default. Examples of this can be found in our own Dockerfiles, which can be found in the [RAPIDS Docker Repository](https://github.com/rapidsai/docker) on GitHub.
```

ajschmidt8: couple of semantic suggestions.

ajschmidt8 (approved):

PR #346, opened 2021-07-12
Remove `xgboost` workaround

ajschmidt8:
This PR removes the temporary workaround for `xgboost` that was added in #243. It can be removed now that https://github.com/dmlc/treelite/issues/244 is closed.


ajschmidt8:
@raydouglass while I'm touching this file, can this conditional block also be removed/consolidated?

https://github.com/rapidsai/docker/blob/87bbdcf5b7ba5f59c6df9ab820a0137f6b6a0fcc/templates/rapidsai-core/Devel.dockerfile.j2#L116-L140

raydouglass:
> @raydouglass while I'm touching this file, can this conditional block also be removed/consolidated?

Yes, I think just keep the `CUDA_VER == 11.0` block.

raydouglass (approved):

Issue #351, opened 2021-07-26
[FEA] Disable Jupyter background launch in runtime notebooks


jacobtomlinson:
Today I ran a Python GPU tutorial for EuroPython. To run the training I launched an instance of [Binder](https://github.com/jupyterhub/binderhub) on GKE which was configured to have one GPU per participant so that they could work through the examples.

Binder can take a GitHub repo and builds it into a Docker image which is then executed by JupyterHub. One way you can configure this is to provide a `Dockerfile` which builds on an existing base image and adds a few binder specific things (mainly around permissions, users, expected workdir and copying in the repo contents).

I was able to use the RAPIDS runtime image as a base and it worked really well 🎉.

One thing I did notice though was that Binder wants to launch Jupyter as the primary command in the container and pass some specific arguments to it. The Jupyter that the RAPIDS image starts as a side effect was interfering with this as there were port collisions.

I couldn't use `JUPYTER_FG` for this as Binder wanted to pass the full jupyter command.

To work around this I had to override the entrypoint script that gets sourced with an empty script to effectively disable this.

https://github.com/jacobtomlinson/gpu-python-tutorial/blob/154bc2c5f562489271c454568c315a71737f5b4c/binder/Dockerfile#L15

Would it be possible to add some kind of flag which would disable Jupyter in the runtime image?

ajschmidt8:
Hey @jacobtomlinson. I'm currently working on this. I'll post a link to the PR here when it's ready.

ajschmidt8:
PR here: https://github.com/rapidsai/docker/pull/356

jacobtomlinson:
Thanks @ajschmidt8

PR #356, opened 2021-08-12
Update `runtime`/`devel` entrypoint

ajschmidt8:
This PR updates the `runtime`/`devel` entrypoint to support a new argument, `DISABLE_JUPYTER`, which disables the default Jupyter server from running when the container starts


ajschmidt8:
Closes #351

Ethyling (approved):

PR #366, opened 2021-09-20
Entrypoint script should error if commands fail

### LINKED ISSUE #361: [BUG] return error codes aren't checked when the extra package installation fails
**Describe the bug**
presently, when we use the RAPIDS nightly image, we specify extra APT/PIP packages etc. to install [thusly](https://github.com/rapidsai/docker/blob/branch-21.10/context/packages.sh#L11). however, when those packages can't be installed for whatever reason, that error code isn't propagate back to the [caller](https://github.com/rapidsai/docker/blob/branch-21.10/context/entrypoint.sh#L14)

the error codes are absorbed and ignored, and the return value if the `entrypoint` script is the return value from [this](https://github.com/rapidsai/docker/blob/branch-21.10/context/entrypoint.sh#L31) command.

   - `entrypoint` invoked like: `DASK_DISTRIBUTED__DASHBOARD__PROMETHEUS__NAMESPACE=ngcjob_appmetrics_dask EXTRA_PIP_PACKAGES=foo /opt/docker/bin/entrypoint --unquote-exec dask-scheduler --dashboard-address :8787 --port 8786 --protocol ws` 

**Expected behavior**
bail out early when the packages couldn't be installed (or provide an option to the user to customize this behavior).

**Environment details (please complete the following information):**
 - Environment location: Docker
 - Method of cuDF install: Docker
   - `docker pull rapidsai/rapidsai-core:21.08-cuda11.0-runtime-ubuntu18.04-py3.7`

**Additional context**
we are invoking this in the NGC context when we create NGC jobs, and want the job creation fail if *any* of the extra packages the user wants couldn't be installed, as the user expects them to be present to successfully run their experiments.

FYI: @jacobtomlinson

jacobtomlinson:
Closes #361

### DIFF
--- a/context/entrypoint.sh
+++ b/context/entrypoint.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 # Activate the `rapids` conda environment.
 . /opt/conda/etc/profile.d/conda.sh
 conda activate rapids


ajschmidt8 (approved):

ajschmidt8:
thanks @jacobtomlinson!!

PR #367, opened 2021-09-20
Add configurable timeouts to package install commands

### LINKED ISSUE #365: [BUG] Bad conda package installs hang container startup
**Describe the bug**
When using the `EXTRA_CONDA_PACKAGES` environment variable conda is invoked by the entrypoint script to install packages before the main process is started. 

If a user provides a package that cannot be resolved by conda (e.g `pytorch`) then this hangs and the container never starts.

```bash
$ docker run --rm -it -e "EXTRA_CONDA_PACKAGES=pytorch" rapidsai/rapidsai-core-nightly:21.10-cuda11.0-runtime-ubuntu18.04 bash
$ # hangs for 30+ minutes
```

You can also reproduce this by starting the container and running `conda install pytorch` manually. 

**Expected behavior**
I would expect package resolution to fail in a timely manner and an error to be returned.

jacobtomlinson:
Adds a timeout to each of the package install commands which default to 10 minutes but can be set with the environment variables `APT_TIMEOUT`, `YUM_TIMEOUT`, `CONDA_TIMEOUT` and `PIP_TIMEOUT`.

Possibly closes #365

### DIFF
--- a/context/packages.sh
+++ b/context/packages.sh
@@ -8,7 +8,7 @@ if [ "$EXTRA_APT_PACKAGES" ]; then
     fi
     echo "EXTRA_APT_PACKAGES environment variable found. Installing packages."
     apt update -y
-    apt install -y --no-install-recommends $EXTRA_APT_PACKAGES
+    timeout ${APT_TIMEOUT:-600} apt install -y --no-install-recommends $EXTRA_APT_PACKAGES
 fi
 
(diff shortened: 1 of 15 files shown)


ajschmidt8 (approved):

jacobtomlinson:
Thanks @ajschmidt8

PR #371, opened 2021-09-24
Update Blazing URL

ajschmidt8:
Since the RAPIDS team has taken over maintenance of Blazing for the `21.10` release, this PR updates the URL that's used to clone Blazing in our `devel` images.

### DIFF
--- a/generated-dockerfiles/rapidsai_centos7-devel.Dockerfile
+++ b/generated-dockerfiles/rapidsai_centos7-devel.Dockerfile
@@ -38,7 +38,7 @@ ENV CUDF_HOME=/rapids/cudf
 
 RUN mkdir -p ${BLAZING_DIR} \
     && cd ${BLAZING_DIR} \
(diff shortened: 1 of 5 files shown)


Ethyling (approved):

generated-dockerfiles/rapidsai_centos7-devel.Dockerfile
```diff
@@ -38,7 +38,7 @@ ENV CUDF_HOME=/rapids/cudf
 
 RUN mkdir -p ${BLAZING_DIR} \
     && cd ${BLAZING_DIR} \
-    && git clone -b ${BUILD_BRANCH} https://github.com/BlazingDB/blazingsql.git
+    && git clone -b ${BUILD_BRANCH} https://github.com/rapidsai/blazingsql-release-staging/.git
```
Ethyling, line 41: We should remove the last `/`

ajschmidt8: good catch! fixed in latest commit.

PR #374, opened 2021-09-30
Update `UCX_PY_VER` Value

ajschmidt8:
This PR updates the `UCX_PY_VER` value for `devel` containers. The `ucx-py` version used for `branch-21.10` should be `0.22` instead of `0.21`. See https://github.com/rapidsai/cudf/pull/9099 for reference.


msadang (approved):
Lgtm

PR #377, opened 2021-10-05
Add arm64 images

Ethyling:


templates/rapidsai-core/partials/install_rapids_env_pkg.j2
```diff
@@ -0,0 +1,18 @@
+{% if "amd64" in arch %}
```
ajschmidt8, line 1: Is this logic only being used in `templates/rapidsai-core/Devel.dockerfile.j2`? If so, you can probably just in-line it. I try to save partials for code that is going to be used across multiple templates.

Ethyling:
> since this is going into `branch-21.12` now, you'll need to update the axes files to use `21.12` instead of `21.10`

We are planning to do a arm release for `21.10`. Should we target `branch-21.10` then for this PR?

ajschmidt8 (approved):

PR #391, opened 2021-10-15
Set entrypoint path as /opt/conda/bin/tini

Ethyling:
As a follow up to rapidsai/gpuci-build-environment#215, we need to update the entrpoint path to /opt/conda/bin/tini


Ethyling:
Closed in favor of https://github.com/rapidsai/docker/pull/392

PR #397, opened 2021-11-04
Add dask-sql to rapidsai images

Ethyling:
Add `dask-sql` to `rapidsai` images


generated-dockerfiles/rapidsai_centos7-devel.amd64.Dockerfile
```diff
@@ -15,49 +15,36 @@ ARG FROM_IMAGE=rapidsai/rapidsai-core-dev
...
 ENV LD_LIBRARY_PATH_ORIG=${LD_LIBRARY_PATH}
 ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/cuda/compat
 
-RUN rm -f ${GCC9_DIR}/lib64/libm.so.6
-
 RUN source activate rapids \
-    && cd ${BLAZING_DIR}/blazingsql \
-    && ./build.sh
+    && cd ${DASK_SQL_DIR}/dask-sql \
+    && python -m pip install . --no-deps -vv
```
charlesbluca, line 38: Do we also need to run `python setup.py java`, or is the Java installation covered here with the `pip install`?

templates/rapidsai/partials/clone_and_build_dask-sql.dockerfile.j2
```diff
@@ -0,0 +1,29 @@
...
+build is done to keep the original LD_LIBRARY_PATH intact. #}
+ENV LD_LIBRARY_PATH_ORIG=${LD_LIBRARY_PATH}
+ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/cuda/compat
+
+RUN source activate rapids \
+    && cd ${DASK_SQL_DIR}/dask-sql \
+    && python -m pip install . --no-deps -vv
+
+ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH_ORIG}
+ENV LD_LIBRARY_PATH_ORIG=
```
ajschmidt8, line 29: I think these `LD_LIBRARY_PATH` changes were all specific to Blazing. So we can probably try removing all references at this point

ajschmidt8 (approved):

generated-dockerfiles/rapidsai_centos7-devel.amd64.Dockerfile
Ethyling: Tbh I don't know, I checked how to build `dask-sql` here: https://github.com/dask-contrib/dask-sql/blob/5b8f8a927fbd8416e52c154f6b4f680b0553fd55/continuous_integration/recipe/meta.yaml#L17
Please let me know if we need to do something else

templates/rapidsai/partials/clone_and_build_dask-sql.dockerfile.j2
Ethyling: Done

generated-dockerfiles/rapidsai_centos7-devel.amd64.Dockerfile
charlesbluca: Ah yeah forgot about that - yeah that seems to trigger the Java build for the nightlies so we should be good here 😄

PR #414, opened 2021-12-15
Fix `clx` images

ajschmidt8:
**Since there are a lot of automatically generated changes in this PR, it's probably easiest to review by commit.**

This PR includes the following changes:

- Removes the `clx` symlink and `ci` folder from `core` images since the `core` images don't have `clx` installed and the `ci` directory is not relevant to our images. This change prevents these folders from showing up in the JupyterLab directory in the browser.
- Removes `WORKDIR` step from `clx` images so that they start in the same directory as the `core` images
- Updates `clx` images to use existing partial for `conda clean` commands
- Adds a step to the `clx` images to re-symlink the `clx` repo in the notebooks directory which was removed in the `core` images.
- Inlines and deletes the `clone_and_build_clx` partial since it's only used in one spot.

The `generated_files` GH check is expected to fail on this PR.


ajschmidt8:
I will test these changes locally in the morning before merging.

Ethyling (approved):

ajschmidt8:
Tested locally. Merging.

PR #424, opened 2022-01-26
Fix package specifications

ajschmidt8:
After https://github.com/rapidsai/integration/pull/411 was merged, the build string for our meta-packages changed from `cuda11.5*` to just `cuda11*`. This change caused our Docker builds to stop working.

Since we now build with CUDA Enhanced Compatibility, I think it should be fine to drop the `=cuda` spec from the strings below. However, an alternative would be to include the `cudatoolkit` spec as shown below, but I don't think that's necessary. 

```sh
gpuci_mamba_retry install -n rapids \
  "rapids-notebook-env=${RAPIDS_VER}" \
  "cudatoolkit=${CUDA_VER}*" \
  && gpuci_conda_retry remove -y -n rapids --force-remove \
  "rapids-notebook-env=${RAPIDS_VER}*"
```

@Ethyling, any thoughts here?

### DIFF
--- a/generated-dockerfiles/rapidsai-core_centos7-base.amd64.Dockerfile
+++ b/generated-dockerfiles/rapidsai-core_centos7-base.amd64.Dockerfile
@@ -38,7 +38,7 @@ RUN source activate rapids \
   && conda config --show-sources \
   && conda list --show-channel-urls
 RUN gpuci_mamba_retry install -y -n rapids \
-  "rapids=${RAPIDS_VER}*=cuda${CUDA_VER}*"
+  "rapids=${RAPIDS_VER}*"
 
 
 RUN source activate rapids \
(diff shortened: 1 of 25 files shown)


Ethyling:
> After [rapidsai/integration#411](https://github.com/rapidsai/integration/pull/411) was merged, the build string for our meta-packages changed from `cuda11.5*` to just `cuda11*`. This change caused our Docker builds to stop working.
> 
> Since we now build with CUDA Enhanced Compatibility, I think it should be fine to drop the `=cuda` spec from the strings below. However, an alternative would be to include the `cudatoolkit` spec as shown below, but I don't think that's necessary.
> 
> ```shell
> gpuci_mamba_retry install -n rapids \
>   "rapids-notebook-env=${RAPIDS_VER}" \
>   "cudatoolkit=${CUDA_VER}*" \
>   && gpuci_conda_retry remove -y -n rapids --force-remove \
>   "rapids-notebook-env=${RAPIDS_VER}*"
> ```
> 
> @Ethyling, any thoughts here?

I think we can drop it for now, and add it later if needed

Ethyling (approved):

PR #432, opened 2022-02-10
Update dask-sql build commands

charlesbluca:
Once https://github.com/dask-contrib/dask-sql/pull/396 is merged in, the build commands for dask-sql will change so that the Java extensions must be built separately from the `pip` install. This PR updates the build commands for the `devel` images accordingly.


ajschmidt8:
@charlesbluca, I've subscribed to that PR so that I will be notified when it merges. Then we can merge this PR.

charlesbluca:
Just merged in - this should be safe to merge

ajschmidt8 (approved):

PR #435, opened 2022-02-26
Update dask-sql build command for `devel` images

charlesbluca:
With https://github.com/dask-contrib/dask-sql/pull/406, we should be able to consolidate the build command for dask-sql so that we no longer need to run `python setup.py java` for a source installation.

Needs https://github.com/dask-contrib/dask-sql/pull/406


ajschmidt8 (approved):
Looks good to me. We'll wait to merge until the upstream PR is merged.

ajschmidt8:
upstream is merged. merging this PR.

PR #436, opened 2022-02-28
Add `libcugraphops`

ajschmidt8:
The `libcugraphops` package is necessary for `cugraph` to build. `libcugraphops` is a private repo though, so this just installs it via `conda` which is similar to what we do for `cumlprims`.


msadang (approved):
Addition of graphops library looks good to me.

sevagh (approved):

PR #440, opened 2022-03-22
Remove `~/.m2` directory after `dask-sql` source build

ajschmidt8:
This PR removes the `~/.m2` directory from the images after the `dask-sql` source build. This change is necessary because the `~/.m2` directory contains `log4j`, which triggers some security alerts for our images. As per @jdye64, the `log4j` dependency is only used during the compile steps (and is not in the resulting `DaskSQL.jar` file) and is therefore safe to simply remove with `rm -rf ~/.m2`.


sevagh (approved):

jdye64 (approved):
Looks good. Nothing in the .m2 repository is needed after the build so this operation has no side effect. Also good that it is in the same RUN command so the offending jars shouldn't be persisted to any layer. LGTM

jolorunyomi (approved):

Issue #439, opened 2022-03-21
[FEA] Install RAPIDS libraries in the default conda environment


daxiongshu:
**Is your feature request related to a problem? Please describe.**
Currently, RAPIDS libraries are installed in the `rapids` conda environment instead of the default conda environment `base` in the rapids docker image `rapidsai/rapidsai-core:22.02-cuda11.5-base-ubuntu20.04-py3.8`. This creates a problem when the default environment `base` is forced to be used or there is no way to activate `rapids` environment when using the docker image. A real use case is [viash](https://github.com/openproblems-bio/neurips2021_multimodal_topmethods/issues/19)

**Describe the solution you'd like**
RAPIDS libraries are installed in the default `base` conda environment. For example, `PyTorch` is installed in the `base` environment in PyTorch docker image, `pytorch/pytorch:1.9.0-cuda11.1-cudnn8-runtime`

daxiongshu:
A related issue might be issue #438

ajschmidt8:
@daxiongshu, thanks for bringing this to our attention. We are currently exploring removing the `rapids` conda environment and using the default `base` conda environment instead. This will require a lot of changes to our images and the associated CI, so there is no ETA at this moment.

raydouglass:
This will be resolved as part of the overhaul: #539 & #545

Issue #438, opened 2022-03-17
[BUG] Environment not set correctly when using `docker exec`


jacobtomlinson:
**Describe the bug**
When using `docker exec` with inline bash commands the environment and path is not set correctly.

**Steps/Code to reproduce bug**

```console
$ # Works
$ docker run -it --rm rapidsai/rapidsai-core:22.02-cuda11.5-runtime-ubuntu20.04-py3.8 /bin/bash -c "which dask-scheduler"
/opt/conda/envs/rapids/bin/dask-scheduler


$ # Works
$ docker run -it --rm rapidsai/rapidsai-core:22.02-cuda11.5-runtime-ubuntu20.04-py3.8 /bin/bash 
(rapids) root@cba78a78c0c4:/rapids/notebooks# which dask-scheduler
/opt/conda/envs/rapids/bin/dask-scheduler


$ # Works
$ docker run -d --name raptest rapidsai/rapidsai-core:22.02-cuda11.5-runtime-ubuntu20.04-py3.8
$ docker exec -it raptest /bin/bash
(rapids) root@079b262f3c0b:/rapids/notebooks# which dask-scheduler
/opt/conda/envs/rapids/bin/dask-scheduler
(rapids) root@079b262f3c0b:/rapids/notebooks# exit
$ docker rm -f raptest


$ # Doesn't work
$ docker run -d --name raptest rapidsai/rapidsai-core:22.02-cuda11.5-runtime-ubuntu20.04-py3.8
$ docker exec -it raptest /bin/bash -c "which dask-scheduler"

$ docker rm -f raptest
```

**Additional context**
The exec is being used by an external health check to ensure Dask is installed in the container. I need to be able to run a one-line exec that verifies that `dask-scheduler` can be called. I can work around by setting the external health check to use the full path, this means the health check is no longer portable to other images.

It seems that inline bash commands are not picking up the environment that is set by the entrypoint script. Perhaps a solution would be for things to be installed in the base environment, or for the rapids environment to be the default, without the entrypoint needing to be called.

PR #449, opened 2022-04-06
Set the default PATH to include the RAPIDS conda env

jacobtomlinson:
As a workaround for issues raised in #438 and #439 this PR sets the `PATH` explicitly to include the RAPIDS conda environment.

Many container platforms including Kubernetes override the default entrypoint, so this change ensures binaries like `jupyter` are still on the path even without the entrypoint being run. Also when calling `docker exec` on a container the environment is not copied from the main process and without the entrypoint the path is also wrong.

### DIFF
--- a/generated-dockerfiles/rapidsai-core_centos7-base.amd64.Dockerfile
+++ b/generated-dockerfiles/rapidsai-core_centos7-base.amd64.Dockerfile
@@ -58,6 +58,9 @@ WORKDIR ${RAPIDS_DIR}
 
 COPY NVIDIA_Deep_Learning_Container_License.pdf . 
 COPY entrypoint.sh /opt/docker/bin/entrypoint
+
+ENV "PATH=/opt/conda/envs/rapids/bin:/opt/conda/condabin:/opt/conda/bin:/usr/local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+
 ENTRYPOINT [ "/opt/conda/bin/tini", "--", "/opt/docker/bin/entrypoint" ]
 
 CMD [ "/bin/bash" ]
--- a/generated-dockerfiles/rapidsai-core_centos7-devel.amd64.Dockerfile
+++ b/generated-dockerfiles/rapidsai-core_centos7-devel.amd64.Dockerfile
@@ -234,6 +234,9 @@ RUN chmod -R ugo+w /opt/conda ${RAPIDS_DIR} \
 COPY NVIDIA_Deep_Learning_Container_License.pdf . 
 COPY source_entrypoints/runtime_devel.sh /opt/docker/bin/entrypoint_source
 COPY entrypoint.sh /opt/docker/bin/entrypoint
+
+ENV "PATH=/opt/conda/envs/rapids/bin:/opt/conda/condabin:/opt/conda/bin:/usr/local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+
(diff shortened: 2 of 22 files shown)


beckernick:
Since this puts the RAPIDS environment at the front of PATH, I believe this is technically a breaking change in some situation due to PATH ordering possibly resulting in different executables being run.

Are breaking changes in containers handled similarly to libraries?

jacobtomlinson:
The `PATH` here is exactly what gets set by `conda activate` in the entry point (I just copy/pasted it). And if the entry point is called then this will be overwritten anyway. So I don't think this is breaking. It just sets it for situations where the entry point isn't called.

ajschmidt8 (approved):
Editing the `PATH` variable is never ideal, but from some offline discussions with @jacobtomlinson, it seems that this is necessary to unblock some folks at Google. So we can go ahead and merge this workaround until we're able to get a proper fix in place.

jacobtomlinson:
Thanks @ajschmidt8. Really appreciate the compromise here.

mmccarty:
Thanks @ajschmidt8

generated-dockerfiles/rapidsai-core_centos7-base.amd64.Dockerfile
```diff
@@ -58,6 +58,9 @@ WORKDIR ${RAPIDS_DIR}
 
 COPY NVIDIA_Deep_Learning_Container_License.pdf . 
 COPY entrypoint.sh /opt/docker/bin/entrypoint
+
+ENV "PATH=/opt/conda/envs/rapids/bin:/opt/conda/condabin:/opt/conda/bin:/usr/local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```
jakirkham, line 62: We do something very similar in conda-forge. Though we do this in [the entrypoint script]( https://github.com/conda-forge/docker-images/blob/a4ecdc0d0e049dc121fa7f59ed97ec88c7238408/scripts/entrypoint#L14 )

jacobtomlinson: Yeah the challenge that we are trying to deal with is that many container platforms do not call the entrypoint script.

jakirkham: Right had a similar discussion with @mmccarty offline

IIUC some call `docker run`, but may `docker exec` into the container as well. With `docker exec`, it is possible to prefix any command with the entrypoint script as part of the command, which would handle this case

Manipulating the `PATH` will largely work in most cases. Where it falls short is `conda activate` injects things into the current shell, which is lost without activation. Also any packages that have activation scripts won't be run. Usually this isn't a big deal, but sometimes it can cause issues

While I agree adding `PATH` was the right thing to do in the short-term, in the medium term we may want to improve on this further to ensure activation is used in these other cases. Solutions could range from having folks run the entrypoint and including that in docs to other fixes in the container that aid with activation. Not entirely sure all of the constraints, but maybe we can discuss those offline

jacobtomlinson: Sure I totally agree. I guess the issue is that the entrypoint is specifically a Docker thing, and there are many other container runtimes and platforms that folks are trying to use our images on (contained, cri-o, kubernetes, kubeflow, sagemaker, vertex, azureml, etc). Most do not call the entrypoint at all if a custom command is set, and many set custom commands that cannot be overridden. So we should try not to rely on the entrypoint as a way to correctly set up our environment.

I think this would mean we need to set things up in the base environment, or manually set the environment up to be in a "post conda activate` state when the container is started. Setting environment variables should be straight forwards, but I hadn't considered activation scripts. 

Let's chat in the team meeting today and we can feedback here or in #438/#439.

PR #460, opened 2022-04-08
Fix `PATH` variable

ajschmidt8:
This PR updates the changes made in #449 to re-order the `PATH` values. Before this change, there were some issues occurring on the `devel` images where `conda` was assuming the base environment was `/opt/conda/envs/rapids` instead of `/opt/conda`.


ajschmidt8:
checks are expected to fail on this PR

raydouglass (approved):

PR #497, opened 2022-09-30
Update `raft` build command

ajschmidt8:
Since https://github.com/rapidsai/raft/pull/801 was merged, we need to update our build commands accordingly for `raft`.


raydouglass (approved):

msadang (approved):
Raft-dash updates look good.

PR #502, opened 2022-10-10
Remove mailcap.py patch for cve-2021-28861

msadang:
Remove mailcap.py patch for cve-2021-28861 and update generated dockerfiles

### DIFF
--- a/templates/rapidsai-core/partials/patch.dockerfile.j2
+++ b/templates/rapidsai-core/partials/patch.dockerfile.j2
@@ -1,7 +1,8 @@
 {# This partial is used to install patches to resolve known CVEs #}
 
 {# Patch for CVE-2021-28861 https://nvd.nist.govb/vuln/detail/CVE-2021-28861 #}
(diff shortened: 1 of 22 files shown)


templates/rapidsai-core/partials/patch.dockerfile.j2
```diff
@@ -1,10 +1,8 @@
 {# This partial is used to install patches to resolve known CVEs #}
 
-{# Patch for CVE-2020-8116 https://github.com/advisories/GHSA-ff7x-qrg7-qggm #}
-{# Patch for CVE-2020-15123 https://nvd.nist.gov/vuln/detail/CVE-2020-15123 #}
-RUN source activate rapids \
```
raydouglass, line 5: Why was this removed? Isn't it still necessary?

msadang: I removed it because, if I remember correctly it was related to the 22.08 release.  I can leave it in.

raydouglass (approved):

PR #515, opened 2022-12-14
Pin pyarrow to 9.0.0

AyodeAwe:
This pin prevents a forced transitive dependency upgrade (Open SSL).


templates/rapidsai-core/partials/install_notebooks.dockerfile.j2
```diff
@@ -3,7 +3,7 @@
 {# Install the rapids-notebook-env meta-pkg #}
 ARG OPENSSL_VERSION=1.1.1
```
ajschmidt8, line 4:
```suggestion
```

This line should also be removed. In the future, consider using `git revert <hash>` to do a complete revert of previous commits instead of manually undoing them.

The Dockerfiles will need to be regenerated again after making this change.

ajschmidt8 (approved):

PR #541, opened 2023-04-03
Refactor dask-sql updating workflow to use shortened version for `DASK_SQL_VER`

charlesbluca:
As discussed a while ago with @ajschmidt8, we are erroneously installing the stable version of Dask-SQL in our nightly containers.

This PR makes some modifications to the version specification and workflow updating these specifications so that we will now properly install the nightly version in these containers as expected.


ajschmidt8 (approved):

ajschmidt8:
thanks Charles!

PR #542, opened 2023-04-05
Use full version for dask-sql

raydouglass:
Follow up from #541 

Since the devel build uses the `DASK_SQL_VER` to pick which branch or tag to clone, it must use the full version to successfully clone https://github.com/dask-contrib/dask-sql/tree/2023.2.0

Also strips the patch version from `DASK_SQL_VER` when installing the dependencies via `mamba`.

See a recent failed build [here](https://gpuci.gpuopenanalytics.com/job/rapidsai/job/docker/job/rapidsai-devel/797/BUILD_IMAGE=rapidsai%2Frapidsai-dev-nightly,CUDA_VER=11.8,DASK_SQL_VER=2023.2,FROM_IMAGE=rapidsai%2Frapidsai-core-dev-nightly,IMAGE_TYPE=devel,LINUX_VER=ubuntu22.04,PYTHON_VER=3.10,RAPIDS_VER=23.04/console)

### DIFF
--- a/ci/axis/rapidsai-devel.yaml
+++ b/ci/axis/rapidsai-devel.yaml
@@ -12,7 +12,7 @@ RAPIDS_VER:
   - '23.04'
 
 DASK_SQL_VER:
-  - '2023.2'
+  - '2023.2.0'
 
 CUDA_VER:
   - 11.8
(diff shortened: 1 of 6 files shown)


charlesbluca (approved):

ajschmidt8 (approved):
LGTM.

The `%.*` expression may only work if our images inherit the `SHELL` directive [from our base images](https://github.com/search?q=repo%3Arapidsai%2Fgpuci-build-environment%20shell&type=code) which is set to `bash` (I'm not sure if Docker inherits this).

Otherwise `RUN` will default to `sh` ([src](https://docs.docker.com/engine/reference/builder/#run)) which may not be compatible with the `%.*` expression.

I haven't tested any of this.

raydouglass:
I ran a test build on my local machine and it works: `#6 60.15 Looking for: ['dask-sql=2023.2']`

Issue #504, opened 2022-10-14
[QST] Can we remove old tags


jacobtomlinson:
**What is your question?**

I noticed a user today trying to use `rapidsai/rapidsai:latest`. Could we clean up old tags that we are no longer pushing to to avoid this confusion?

raydouglass:
As part of the overhaul #539 & #545, I will be removing old tags after the `23.08` release.

PR #545, opened 2023-04-17
`docker` repo overhaul

### LINKED ISSUE #539: Repository overhaul for RAPIDS v23.08
# `docker` repo (RAPIDS end-user images) overhaul

This issue outlines how the RAPIDS end-user images in the `docker` repo will be overhauled to improve functionality and performance.

### Requirements

These are the requirements for all images produced by this repo.

1. Must use multi-architectural images that publish `amd64` and `aarch64`/`arm64` images to the same docker repository
    a. This eliminates the `arm64` variant repositories such as https://hub.docker.com/r/rapidsai/rapidsai-core-arm64
2. Must install all conda package into the `base` conda environment
3. Default user must be non-`root` user named `rapids`
4. Must define the `WORKDIR` as the `rapids` user's home directory
6. Must use a tag in the format: `${RAPIDS_VER}-cuda${CUDA_VER}-py${PYTHON_VER}`
7. Must support the current matrix of CUDA and Python versions (currently defined [here](https://github.com/rapidsai/docker/blob/branch-23.04/ci/axis/rapidsai-core-base-notebooks.yaml))

### Considerations

1. Consider defining any dependencies not inherited from RAPIDS libraries in a `dependencies.yaml` consumable by [`dependency-file-generator`](https://github.com/rapidsai/dependency-file-generator)
  a. If there aren't many, might be simpler to directly install them (eg `mamba install rapids ipython`)
2. Consider implementing `sudoers` allowing the `rapids` user to `sudo`

### Base image

This image is the basic docker image that installs the RAPIDS suite of libraries.

#### Requirements

1. Must `FROM` [`rapidsai/mambaforge-cuda`](https://github.com/rapidsai/mambaforge-cuda)
2. Must use default `CMD` to launch an `ipython` REPL
3. Must install the [`rapids` meta package](https://github.com/rapidsai/integration/tree/branch-23.04/conda/recipes/rapids)
4. Must use `dask`/`distributed`/`dask-sql` nightly conda builds for nightly image builds
5. Must include `ipython` dependency
6. Must be published to `rapidsai/rapidsai`

### Notebooks image

This image extends the base image to include [Jupyter](https://jupyter.org/) and [RAPIDS notebooks](https://github.com/rapidsai/notebooks).

#### Requirements

1. Must `FROM` the base image define above
2. Must use default `CMD` to launch (in the foreground) a jupyter server
3. Must install the required "notebook" dependencies from all RAPIDS repositories (eg https://github.com/rapidsai/cudf/blob/branch-23.04/dependencies.yaml#L304)
4. Must clone the [RAPIDS notebooks](https://github.com/rapidsai/notebooks) into the `WORKDIR`
5. Must be published to `rapidsai/rapidsai-notebooks`

## Implementation

Building the base and notebooks images should utilize a multi-stage docker build.

### Stage 1

1. Clone RAPIDS repositories for the current branch
2. Generate conda environment files for each repositories' "notebook" dependencies
3. Merge the conda environment files
4. Collect the `notebooks` directory of each RAPIDS repository and associated datasets if required

### Stage 2

1. Build the base image defined above

### Stage 3

1. Pull the merged conda environment file from Stage 1
2. Copy the notebooks directories from Stage 1
3. Build the notebooks image defined above using the merged conda environment file for dependencies

### Base image

The base image will be tested using basic RAPIDS smoke tests being developed in their respective repositories. The full suite of unit tests will not be run.

### Notebooks image

The notebooks image will be tested using the "notebook tests" which basically executes the notebooks provided by each repositories (consolidated in the [`notebooks` repo](https://github.com/rapidsai/notebooks)).

It will be necessary to standardize the mechanism for testing each notebook and for exempting a notebook from testing. `cugraph` provides a reliable way for doing this: https://github.com/rapidsai/cugraph/blob/branch-23.04/ci/notebook_list.py. 

## Breaking changes

These are the breaking changes that will occur when v23.06 of this `docker` repository is released:

1. `rapidsai/rapidsai-core` is deprecated and superceded by `rapidsai/rapidsai` and `rapidsai/rapidsai-notebooks`
2. "Runtime" images (eg `rapidsai/rapidsai:23.02-cuda11.8-runtime-ubuntu22.04-py3.10`) are replaced by the `rapidsai/rapidsai-notebooks` docker repo
3. `IMAGE_TYPE` will be removed from image tags. Change `${RAPIDS_VER}-cuda${CUDA_VER}-${IMAGE_TYPE}-${LINUX_VER}-py${PYTHON_VER}` to `${RAPIDS_VER}-cuda${CUDA_VER}-${LINUX_VER}-py${PYTHON_VER}`
4. `arm64` image repositories (eg `rapidsai/rapidsai-arm64`) are deprecated and superceded by multi-architectural images in `rapidsai/rapidsai` and `rapidsai/rapidsai-notebooks`
5. Images will not longer default to the `root` user. Instead they will use non-root user `rapids`
  a. To use previous behavior, supply `--user root` to the `docker run` command
6. The `base` image will start an `ipython` REPL instead of a `bash` shell
  a. To use previous behavior, supply a `docker run` `CMD` of `bash` (eg `docker run -it ${IMAGE} bash`)
7. The `notebooks` image will start `jupyter-lab` in the foreground instead of running in the background and starting a `bash` shell.
  a. To use previous behavior, supply a `docker run` `CMD` of `bash` (eg `docker run -it ${IMAGE} bash`) and run `jupyter-lab` in the [background](https://bash.cyberciti.biz/guide/Putting_jobs_in_background)
8. The notebooks directory in the `notebooks` image will not longer contain the full source code git clone of RAPIDS repositories
  a. To use previous behavior, start the container and `git clone ${RAPIDS_GIT_URL}` for each required RAPIDS repository
9. The `rapids` conda environment is removed and replaced by the `base` conda environment
10. The environment variables `EXTRA_APT_PACKAGES` & `EXTRA_YUM_PACKAGES` will not longer be supported.
11. The devel images will be removed and replaced by devcontainers (such as https://github.com/rapidsai/cuspatial/pull/960 for example)
12. `LINUX_VER` will be removed from the image tags. The Ubuntu version is implied by the CUDA version.

raydouglass:
Closes #539 

This PR overhauls the `docker` according to #539.

Multi-arch builds inspired from https://github.com/rapidsai/mambaforge-cuda/pull/31

Note: this PR merges into `branch-23.08-overhaul` instead of `branch-23.08`. `branch-23.08-overhaul` was manually created to remove the vast majority of the repo. Merging this PR into that branch shows more obvious changes instead of a [long list of deleted files](https://github.com/rapidsai/docker/compare/branch-23.08...branch-23.08-overhaul).

To-do:
- [x] Handle uploading nightlies with alpha version tag
- [x] `main` builds should clean up `condarc`
- [x] Add tests
- [x] Test all notebooks (not just cudf)

Follow up PRs:
- Dockerhub/NGC readme updater
- Enable nightly tests
- Remove required `up-to-date` status check
- Merge `branch-23.08-overhaul` into `branch-23.08`
- Add style checking (https://github.com/rapidsai/docker/pull/545#discussion_r1264520507)


ci/build_and_tag_images.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
ajschmidt8, line 1:
```suggestion
#!/bin/bash
set -euo pipefail
```

We should add some of these flags to catch errors.

We should also consider standardizing which flags we use across all files.

Looks like there are some inconsistencies currently.

ajschmidt8 (commented):
Nice!

This is a good start. I still have to give it a more thorough review.

In the meantime, I noticed some bash flags inconsistencies that we should probably normalize.

ci/build_and_tag_images.sh
raydouglass: I've updated this throughout the repo.

.github/workflows/build-and-publish-images.yml
```diff
@@ -0,0 +1,143 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Compute matrix
+        id: compute-matrix
+        run: ./ci/compute-matrix.sh
+  docker-build-per-arch:
+    needs: compute-matrix
+    strategy:
+      matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
+    runs-on: ${{ fromJSON(matrix.runner_arch != 'arm64' && '"ubuntu-latest"' || '{"labels":["self-hosted", "linux", "arm64", "cpu4"]}') }}
```
jjacobelli, line 56: We need to change `'{"labels":["self-hosted", "linux", "arm64", "cpu4"]}'` into `'linux-arm64-cpu4'` to use ARC V2 runners

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+    && conda clean -afy
+
+ENV DASK_LABEXTENSION__FACTORY__MODULE="dask_cuda"
+ENV DASK_LABEXTENSION__FACTORY__CLASS="LocalCUDACluster"
+
+EXPOSE 8888
+
+ENTRYPOINT ["/home/rapids/entrypoint.sh"]
+
+CMD ["jupyter-lab", "--allow-root", "--notebook-dir=/home/rapids/notebooks", "--ip=0.0.0.0", "--no-browser", "--NotebookApp.token=''", "--NotebookApp.allow_origin='*'"]
```
jjacobelli, line 89: Do we still need the `--allow-root` flag?

raydouglass: Nope

.github/workflows/build-and-publish-images.yml
raydouglass: Updated!

Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+COPY entrypoint.sh /home/rapids/entrypoint.sh
+
+# CI should handle modifying this file instead of the dockerfile
+# RUN if [ "${RAPIDS_BRANCH}" = "main" ]; then sed -i '/nightly/d;/dask\/label\/dev/d' /opt/conda/.condarc; fi
+
+RUN mamba install -y -n base \
+        "rapids=${RAPIDS_VER}.*" \
+        "dask-sql=${DASK_SQL_VER%.*}.*" \
+        "python=${PYTHON_VER}.*" \
+        "cudatoolkit=${CUDA_VER%.*}.*" \
```
jakirkham, line 53: Would suggest switching to the `cuda-version` package. This properly [constrains `cudatoolkit` on CUDA 11]( https://github.com/conda-forge/cuda-version-feedstock/blob/7e8b642c6bb4aeda93938a4cedc13895388937fe/recipe/meta.yaml#L17 ) and [on CUDA 12 this constrains the CTK]( https://github.com/conda-forge/cuda-version-feedstock/tree/main/recipe#readme )

```suggestion
        "cuda-version=${CUDA_VER%.*}.*" \
```

bdice: I think we _also_ need to install `cudatoolkit` on CUDA 11?

jakirkham: Am suggesting the latter, but have no strong preference

raydouglass:
`dask-sql` is looking into supporting nightlies for `arm64`: https://github.com/dask-contrib/dask-sql/pull/1144

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,106 @@
...
+          mkdir -p /tmp/gha-tools
+          curl -s -L 'https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz' | tar -xz -C /tmp/gha-tools
+          echo "/tmp/gha-tools" >> "${GITHUB_PATH}"
+      - name: Clean up condarc for release builds
+        run: |
+          GIT_DESCRIBE_TAG="$(git describe --tags --abbrev=0)"
+          GIT_DESCRIBE_TAG="${GIT_DESCRIBE_TAG:1}" #remove leading 'v'
+          if [[ ! $GIT_DESCRIBE_TAG =~ [a-z] ]]; then
+            rapids-logger "Most recent tag is for release, removing nightly channels"
+            sed -i '/nightly/d;/dask\/label\/dev/d' context/condarc
```
raydouglass, line 59: This is a hacky solution. Open to suggestions for improvements!

context/test_notebooks.py
bdice, line 1: At some point (maybe after this PR) we should add a basic `.pre-commit-config.yaml` to enforce code quality. I'd suggest:
- Python linters from other RAPIDS repos like `black`, `isort`, and perhaps `ruff` instead of `flake8` (there's mixed usage of single quotes and double quotes, atypical import ordering, etc.)
- Repo-wide checks for normalized newlines / EOF newlines
- Dockerfile and YAML linters to prevent syntax errors

.github/workflows/build-image.yml
raydouglass: There's `rapidsai-nightly` & `dask/label/dev`

context/test_notebooks.py
```diff
@@ -0,0 +1,192 @@
...
+import timeit
+from typing import Iterable
+import nbconvert
+import nbformat
+from datetime import datetime
+from nbconvert.preprocessors import ExecutePreprocessor
+import yaml
+
+
+bad_subdirectory_names = [
```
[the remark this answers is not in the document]
raydouglass: Switched to `ignored`

context/test_notebooks.py
raydouglass: Added to the list of follow up PRs in this PR's body.

README.md
```diff
@@ -1,75 +1,50 @@
...
+It extends the `base` images to include RAPIDS notebooks and a [`jupyterlab` server](https://jupyter.org/) which starts automatically.
 
-## Extending Images
+## Image tags
 
-Like any Docker image, the RAPIDS images can be extended to suit the needs of individual teams. Whether it is to add custom libraries, change security settings, or other customizations; using `FROM` and our RAPIDS images allows users to customize the container, but easily update to the latest versions with a new `docker build`.
+Tags for both `base` and `notebooks` images take the form of `${RAPIDS_VER}-cuda${CUDA_VER}-${LINUX_VER}-py${PYTHON_VER}`.
 
-### Custom Token Example
+There is no `latest` tag.
```
ajschmidt8, line 27: Ah. I see now that we don't want to support a `latest` tag.

We should delete the existing `latest` tag or consider adding `latest` tag support to resolve https://github.com/rapidsai/docker/issues/504.

ajschmidt8 (commented):
Awesome job.

Looks good overall.

Left a few comments/questions/suggestions.

context/notebooks.sh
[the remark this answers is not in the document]
ajschmidt8: For reference, a query like this `conda list  --json | jq '.[] | select(.name == "rmm")'` will return an object like this:

```json
{
  "base_url": "https://conda.anaconda.org/rapidsai",
  "build_number": 0,
  "build_string": "cuda11_py310_230607_gacaf3f5e_0",
  "channel": "rapidsai",
  "dist_name": "rmm-23.06.00-cuda11_py310_230607_gacaf3f5e_0",
  "name": "rmm",
  "platform": "linux-64",
  "version": "23.06.00"
}
```

From there we could parse the `build_string` for the `git` SHA.

Though right now we don't enforce that our C++ and Python packages are tied to the same SHA, so this might be a moot point for now.

Maybe we should solve that first and then come back to this issue.

.github/workflows/build-image.yml
taureandyernv: If we're doing ARM testing, what about `rapids-core-nightly-arm64`?

raydouglass: We're talking about conda channels not image types. We want to ensure that a release does not use nightly conda channels.

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Clean up condarc for release builds
+        run: |
+          GIT_DESCRIBE_TAG="$(git describe --tags --abbrev=0)"
+          GIT_DESCRIBE_TAG="${GIT_DESCRIBE_TAG:1}" # remove leading 'v'
+          if [[ ! $GIT_DESCRIBE_TAG =~ [a-z] ]]; then
+            rapids-logger "Most recent tag is for release, removing `rapidsai-nightly` and `dask/label/dev` channels."
+            sed -i '\|rapidsai-nightly|d;\|dask/label/dev|d' context/condarc
+          else
+            rapids-logger "Most recent tag is an alpha. Build will use nightly channels."
+          fi
```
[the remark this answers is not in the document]
raydouglass: Let me give this a test. Perhaps at a minimum `rapidsai-nightly` should be prioritized lower than `rapidsai`?

raydouglass:
> I'm also curious about this.
> I think since the process is running in a container it should be okay to omit 0.0.0.0.
> I would imagine that the docker run arguments would be responsible for determine which interface(s) to listen on.

Not sure why I can't reply inline, but I tested without the `--ip=0.0.0.0` and it does not work. I think this is because without it, jupyter binds to `127.0.0.1` instead of the docker IP assigned (something in `172.17.0.0/16` usually). `--ip=0.0.0.0` causes it to bind to every IP.

.github/workflows/build-test-publish-images.yml
```diff
@@ -0,0 +1,254 @@
...
+      RAPIDS_VER: ${{ needs.compute-matrix.outputs.RAPIDS_VER }}
+      DASK_SQL_VER: ${{ matrix.DASK_SQL_VER }}
+      BASE_TAG:
+        "rapidsai/${{ needs.compute-matrix.outputs.BASE_IMAGE_REPO }}:\
+        ${{ needs.compute-matrix.outputs.BASE_TAG_PREFIX }}\
+        ${{ needs.compute-matrix.outputs.RAPIDS_VER }}\
+        ${{ needs.compute-matrix.outputs.ALPHA_TAG }}-\
+        cuda${{ matrix.CUDA_VER }}-\
+        ${{ matrix.LINUX_VER }}-\
+        py${{ matrix.PYTHON_VER }}"
```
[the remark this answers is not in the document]
raydouglass: I was planning to drop the shortcut tags altogether including `latest`. But if someone has a strong opinion, I'm open to the discussion.

.github/workflows/build-test-publish-images.yml
```diff
@@ -0,0 +1,254 @@
...
+            base_source_tags+=("${base_tag}-${arch}")
+            notebooks_source_tags+=("${notebooks_tag}-${arch}")
+          done
+
+          docker manifest create ${base_tag} ${base_source_tags[@]}
+          docker manifest push ${base_tag}
+          docker manifest create ${notebooks_tag} ${notebooks_source_tags[@]}
+          docker manifest push ${notebooks_tag}
+  delete-temp-images:
+    if: always()
```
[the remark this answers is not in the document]
raydouglass: I based it off this post you made actually: https://github.com/rapidsai/mambaforge-cuda/pull/31/files#r1189940564

.github/workflows/build-test-publish-images.yml
```diff
@@ -0,0 +1,254 @@
...
+          base_tag_array=(
+            rapidsai/${{ needs.compute-matrix.outputs.BASE_IMAGE_REPO }}:
+            ${{ needs.compute-matrix.outputs.BASE_TAG_PREFIX }}
+            ${{ needs.compute-matrix.outputs.RAPIDS_VER }}
+            ${{ needs.compute-matrix.outputs.ALPHA_TAG }}-
+            cuda${{ matrix.CUDA_VER }}-
+            ${{ matrix.LINUX_VER }}-
+            py${{ matrix.PYTHON_VER }}
+          )
+          base_tag="$(printf %s "${base_tag_array[@]}" $'\n')" # Converts array to string w/o spaces
```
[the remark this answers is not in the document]
raydouglass: I made this change based on the comment here: https://github.com/rapidsai/docker/pull/545#discussion_r1264190712

It's just an effort to make it more readable. I would like to find a better way to construct these strings though since they are (mostly) reused several times.

Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+
+USER rapids
+
+WORKDIR /home/rapids
+
+COPY condarc /opt/conda/.condarc
+
+RUN mamba install -y -n base \
+        "rapids=${RAPIDS_VER}.*" \
+        "dask-sql=${DASK_SQL_VER%.*}.*" \
```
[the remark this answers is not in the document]
raydouglass: Yes!

Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+ENV DASK_LABEXTENSION__FACTORY__MODULE="dask_cuda"
+ENV DASK_LABEXTENSION__FACTORY__CLASS="LocalCUDACluster"
+
+COPY test_notebooks.py /home/rapids/
+
+EXPOSE 8888
+
+ENTRYPOINT ["/home/rapids/entrypoint.sh"]
+
+CMD ["jupyter-lab", "--notebook-dir=/home/rapids/notebooks", "--ip=0.0.0.0", "--no-browser", "--NotebookApp.token=''", "--NotebookApp.allow_origin='*'"]
```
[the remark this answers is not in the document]
raydouglass: I tested without the `--ip=0.0.0.0` and it does not work. I think this is because without it, jupyter binds to `127.0.0.1` instead of the docker IP assigned (something in `172.17.0.0/16` usually). `--ip=0.0.0.0` causes it to bind to every IP.

README.md
raydouglass: I plan to delete it after the release.

README.md
```diff
@@ -1,75 +1,50 @@
...
-```
-
-Once built, the resulting image will be secured with the new token.
-
-This example can be repurposed by replacing the `sed` command with other commands for custom libraries or settings.
+- `EXTRA_CONDA_PACKAGES` - used to install additional `conda` packages in the container. Use a space separated list of conda version specs
+- `CONDA_TIMEOUT` - how long (in seconds) the `conda` install should wait before exiting
+- `EXTRA_PIP_PACKAGES` - used to install additional `pip` packages in the container. Use a space separated list of pip version specs
+- `PIP_TIMEOUT` - how long (in seconds) the `pip` install should wait before exiting
+- `UNQUOTE` -
```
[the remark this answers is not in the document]
raydouglass: Whoops, I'll add the documentation.

context/condarc
```diff
@@ -0,0 +1,8 @@
+auto_update_conda: False
+channels:
+  - rapidsai
+  - rapidsai-nightly
+  - dask/label/dev
+  - pytorch
+  - conda-forge
+  - nvidia
```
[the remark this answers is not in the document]
raydouglass: I think since these are end-user images meant for interactivity, we should use default conda behaviors including `number_channel_notices` and `always_yes`.

I know when I'm doing something in the `rapidai/ci` images, the `always_yes: true` trips me up occasionally.

context/notebooks.sh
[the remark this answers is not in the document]
raydouglass: I think this idea a lot.

One potential problem is we wouldn't know the SHAs until after installing the packages. Which means it would be sort of complicated to gather the notebooks in a separate build stage. Might be able to do the install & notebooks gathering in one stage and copy the entire `/opt/conda` dir in the next stage.

.github/workflows/build-image.yml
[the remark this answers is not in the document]
raydouglass: Removing `rapidsai` from a "nightly" seems to have worked just fine. I think I will commit this change and if it becomes problematic we can revert.

.github/workflows/build-test-publish-images.yml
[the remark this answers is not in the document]
ajschmidt8: I'm fine with dropping.

context/notebooks.sh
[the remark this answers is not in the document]
ajschmidt8: Yeah, this would be nice to have.

I think we can file this as a future enhancement for now.

ajschmidt8 (commented):
This is minor, but can we switch from using the term `axis` to `matrix`.

I think `axis` is mostly used from the Jenkins days, but `matrix` is probably more appropriate for GHAs.

e.g. `matrix.yaml` and `matrix-test.yaml` instead of `axis.yaml` and `axis-test.yaml`.

axis.yaml
```diff
@@ -0,0 +1,14 @@
+CUDA_VER:
+  - "11.2.2"
+  - "11.8.0"
+PYTHON_VER:
+  - "3.9"
+  - "3.10"
+LINUX_VER:
+  - "ubuntu20.04"
+  - "ubuntu22.04"
```
ajschmidt8, line 9: Do we need both `ubuntu` versions?

Didn't we talk about just supporting the latest `ubuntu` version and omitting the OS entirely from the resulting tag name?

Logically it seems like we should use a single OS (the latest Ubuntu) for the same reasons that we dropped CentOS/RockyLinux image variants.

raydouglass: Yes we did discuss using a single Ubuntu. Unfortunately, at the time we did not consider that we had also committed to supporting CUDA 11.2.x (eg we only [dropped 11.4 & 11.5](https://docs.rapids.ai/notices/rsn0027/)), but CUDA 11.2 is not supported on Ubuntu 22.04.

So I decided to add Ubuntu 20.04 for the CUDA 11.2 support.

raydouglass: We could build Ubuntu 20.04 only for CUDA 11.2 and Ubuntu 22.04 for CUDA 11.8+. And then drop the `LINUX_VER` from the tag. Basically, the CUDA version you pull dictates the Ubuntu version.

This would be less transparent, but simpler in a way.

ajschmidt8: > We could build Ubuntu 20.04 only for CUDA 11.2 and Ubuntu 22.04 for CUDA 11.8+. And then drop the `LINUX_VER` from the tag. Basically, the CUDA version you pull dictates the Ubuntu version.
> 
> This would be less transparent, but simpler in a way.

I'm in favor of this.

It's less complicated for our end users. They don't need to worry about what Ubuntu version they're pulling, just the CUDA & Python version.

The consequence of not doing this will mean that everyone who consumes our images will have to keep track of which Ubuntu version they need to specify for each CUDA version. It also means that logic will need to be incorporated in other areas across RAPIDS like the release selector on our website.

bdice: I'm weakly in favor of retaining the Ubuntu version in the tag. For folks building on top of RAPIDS containers, it can make a pretty big difference if you have 20.04 vs. 22.04. Having that be implicit in the CUDA version is not ideal and doesn't communicate helpful info to users. I always search/browse tags directly on Docker Hub, so I like to know what I'm getting.

context/condarc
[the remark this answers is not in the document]
bdice: Got it, that makes sense. Let's leave it as-is and not adopt any other non-default settings.

Dockerfile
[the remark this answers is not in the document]
bdice: Sounds good. As I suspected, Docker networking requires this. 🙃

.github/workflows/build-test-publish-images.yml
```diff
@@ -0,0 +1,254 @@
...
+      ALPHA_TAG: ${{ steps.compute-rapids-ver.outputs.ALPHA_TAG }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Compute matrix
+        id: compute-matrix
+        run: |
+          MATRIX=$(ci/compute-mx.sh)
```
bdice, line 51: Can we rename `mx` to `matrix`? I think my previous comment got lost. https://github.com/rapidsai/docker/pull/545/#discussion_r1264517848

bdice (approved):
Looks fine to me!

axis.yaml
ajschmidt8: > I'm weakly in favor of retaining the Ubuntu version in the tag. For folks building on top of RAPIDS containers, it can make a pretty big difference if you have 20.04 vs. 22.04. Having that be implicit in the CUDA version is not ideal and doesn't communicate helpful info to users. I always search/browse tags directly on Docker Hub, so I like to know what I'm getting.

I'd rather us include the mapping between CUDA version & Ubuntu version in the GitHub/DockerHub README's than in the image tags.

Something like this...

```markdown
The Ubuntu version included in each image is the latest supported by each CUDA version.

See the table below for details.

| CUDA Version        | Ubuntu Version       |
| ------------------- | -------------------- |
| <=11.2              | 20.04                |
| >11.2               | 22.04                |
```

IMO, we should only include the items that we explicitly paramterize (Python & CUDA version) in the image tags.

raydouglass: I'm leaning towards the implicit Ubuntu version and documenting it as AJ suggests.

This is not uncommon for docker tags. For example, `python:3.10` is Debian 12 based and an alias of `python:3.10-bookworm` but you could use `python:3.10-bullseye` or `python:3.10-alpine` to get a different base OS. But the default tag essentially picks the "works-best-for-most" OS.

I don't think we want to build the entire matrix of base OSes, but at least there is some precedent for selecting the "best fit" OS automatically.

bdice: Forgive me, I might be missing something here. If I am reading this right, we aren't going to be shipping containers with other base OSes like Rocky Linux or CentOS (only ubuntu20.04 and ubuntu22.04 are listed here) It looks like `rapidsai/rapidsai` has tags for `centos7` and so forth, but the overhauled images won't support RHEL-like images? If we're only shipping Ubuntu, then the calculus changes for me, and I'm okay with making the OS more of a hidden parameter.

ajschmidt8: That's correct, Bradley. We aren't shipping `centos`/`rockylinux` variants for the overhauled images.

Ok, so it seems like we're all on the same page now then.

Let's nix those OS strings.

matrix.yaml
```diff
@@ -0,0 +1,16 @@
+CUDA_VER:
+  - "11.2.2"
+  - "11.8.0"
+PYTHON_VER:
+  - "3.9"
+  - "3.10"
+LINUX_VER:
+  - "ubuntu20.04"
+  - "ubuntu22.04"
```
ajschmidt8, line 9: since `LINUX_VER` is now a computed property and not something that we explicitly paramaterize, what do you think about moving it to the `compute-matrix.jq` function?

This follows what we do for the `ARCHES` values. It would also allow us to clean this file up a bit.

raydouglass: Good idea!

ajschmidt8: > Would be good to remove `DASK_SQL_VER` too since it's not really a matrix parameter. But that can be a future enhancement I think.

I was thinking the same.

but then I noticed that the Dockerfile already has the version included here: https://github.com/raydouglass/docker/blob/32fa035357fa0f0603e7c24861f08000cc0dd8b4/Dockerfile#L8.

So it should be a no-op to just drop it from the matrix and associated GHAs workflows now and use that Dockerfile `ARG` version as the source of truth, right?

bdice (approved):
Re-approving with latest changes since my last review. Let's ship it! 🚢

matrix.yaml
raydouglass: I think the `update-dask-sql.yml` workflow would need to be updated too. But that whole workflow should probably be replaced by renovate which is a bigger job.

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,106 @@
...
+      fail-fast: false
+    runs-on: "linux-${{ matrix.ARCH }}-cpu4"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Install gha-tools
+        run: |
+          mkdir -p /tmp/gha-tools
```
vyasr, line 50: Is there any upstream image we could use to do the build that would already contain gha-tools?

.github/workflows/build-test-publish-images.yml
```diff
@@ -0,0 +1,246 @@
...
+      NOTEBOOKS_TAG:
+        "rapidsai/${{ needs.compute-matrix.outputs.NOTEBOOKS_IMAGE_REPO }}:\
+        ${{ needs.compute-matrix.outputs.NOTEBOOKS_TAG_PREFIX }}\
+        ${{ needs.compute-matrix.outputs.RAPIDS_VER }}\
+        ${{ needs.compute-matrix.outputs.ALPHA_TAG }}-\
+        cuda${{ matrix.CUDA_VER }}-\
+        py${{ matrix.PYTHON_VER }}"
+  test:
+    needs: [build, compute-matrix]
+    # TODO: nightly tests
```
vyasr, line 126: IIUC right now we only have `pull-request` (pr.yml, which builds and tests) and `branch` (publish.yml, which builds and publishes). So in addition to updating this bit of this workflow, we'll also need to add an additional new workflow for nightly tests. Should we track that work in a new issue?

.github/workflows/test-image.yml
vyasr, line 1: Can we rename this file to `test-notebooks` if it is specific to notebooks only (and not, for instance, running the unit tests in each package)?

.github/workflows/test-image.yml
```diff
@@ -0,0 +1,98 @@
...
+      - name: Print environment
+        run: |
+          rapids-print-env
+          rapids-logger "nvidia-smi"
+          nvidia-smi
+      - name: Test notebooks
+        run: /home/rapids/test_notebooks.py -i /home/rapids/notebooks -o /home/rapids/notebooks_output
+      - name: Install awscli
+        if: '!cancelled()'
+        run: rapids-mamba-retry install -n base awscli
```
vyasr, line 89: Not in scope for this PR, but it's worth noting that awscli v2 is out now and we may want to coordinate an upgrade in all our images.

Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+
+COPY entrypoint.sh /home/rapids/entrypoint.sh
+
+ENTRYPOINT ["/home/rapids/entrypoint.sh"]
+
+CMD ["ipython"]
+
+
+# Notebooks image
+FROM base as notebooks
```
vyasr, line 60: Do we like having multiple images in a Dockerfile? I rarely see it done, although I do see some advantages here.

ci/release/update-version.sh
```diff
@@ -0,0 +1,35 @@
...
+CURRENT_MAJOR=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[1]}')
+CURRENT_MINOR=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[2]}')
+CURRENT_PATCH=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[3]}')
+CURRENT_SHORT_TAG=${CURRENT_MAJOR}.${CURRENT_MINOR}
+
+#Get <major>.<minor> for next version
+NEXT_MAJOR=$(echo $NEXT_FULL_TAG | awk '{split($0, a, "."); print a[1]}')
+NEXT_MINOR=$(echo $NEXT_FULL_TAG | awk '{split($0, a, "."); print a[2]}')
+NEXT_SHORT_TAG=${NEXT_MAJOR}.${NEXT_MINOR}
+# NEXT_UCX_PY_VERSION="$(curl -sL https://version.gpuci.io/rapids/${NEXT_SHORT_TAG}).*"
```
vyasr, line 21:
```suggestion
```

context/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+cat << EOF
+This container image and its contents are governed by the NVIDIA Deep Learning Container License.
+By pulling and using the container, you accept the terms and conditions of this license:
+https://developer.download.nvidia.com/licenses/NVIDIA_Deep_Learning_Container_License.pdf
+
+EOF
+
+if [ -e "/home/rapids/environment.yml" ]; then
+    echo "environment.yml found. Installing packages"
+    timeout ${CONDA_TIMEOUT:-600} mamba env update -n base -f /home/rapids/environment.yml || exit $?
```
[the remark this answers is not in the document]
vyasr: I'm OK waiting to do that in a follow-up

context/notebooks.sh
```diff
@@ -0,0 +1,29 @@
...
+        rapids-dependency-file-generator \
+            --config "$REPO/dependencies.yaml" \
+            --file_key test_notebooks \
+            --matrix "cuda=${CUDA_VER%.*};arch=$(arch);py=${PYTHON_VER}" \
+            --output conda > "/dependencies/${REPO}_notebooks_tests_dependencies.yaml"; \
+    fi
+done
+
+pushd "/dependencies"
+conda-merge ./*.yaml > /test_notebooks_dependencies.yaml
```
vyasr, line 29: I'm a bit surprised that this script doesn't actually _install_ the dependencies. How are people expected to use the notebooks image? It looks like running this script is the last thing that happens in the Dockerfile.

vyasr (approved):
Some minor thoughts, but nothing worth blocking over. Thanks Ray!

.github/workflows/test-image.yml
```diff
@@ -0,0 +1,98 @@
...
+      - uses: aws-actions/configure-aws-credentials@v1-node16
+        if: '!cancelled()'
+        with:
+          role-to-assume: ${{ vars.AWS_ROLE_ARN }}
+          aws-region: ${{ vars.AWS_REGION }}
+          role-duration-seconds: 43200 # 12h
+      - name: Upload notebook test outputs
+        if: '!cancelled()'
+        run: rapids-upload-to-s3 notebooks_output.tar.gz /home/rapids/notebooks_output
```
ajschmidt8, line 98: This is a matrix job, but each matrix instance uploads the same file. I would expect that these uploads will clobber each other unless they include matrix-specific information in their filename.

Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+
+# Base image
+FROM ${BASE_FROM_IMAGE}:cuda${CUDA_VER}-base-${LINUX_VER}-py${PYTHON_VER} as base
+ARG CUDA_VER
+ARG PYTHON_VER
+
+ARG RAPIDS_VER
+ARG DASK_SQL_VER
+
+RUN useradd -rm -d /home/rapids -s /bin/bash -g conda -u 1001 rapids
```
ajschmidt8, line 36:
```suggestion
RUN useradd -rm -d /home/rapids -s /bin/bash -g conda -u 1000 rapids
```

Is there anything preventing us from using `1000` here?

This would match the `id` used for the first user in OSes like Ubuntu.

.github/workflows/build-test-publish-images.yml
raydouglass: Yes, there are a several follow ups to this PR that I intend to open issues and/or PRs for.

.github/workflows/test-image.yml
raydouglass: Yes

.github/workflows/test-image.yml
raydouglass: I will file an issue for this.

context/notebooks.sh
raydouglass: There are three stages to the build: dependencies (which runs this scripts), base, & notebooks. The notebooks stage depends on both base & dependencies. It grabs the resulting `test_notebooks_dependencies.yaml` from the dependencies stage and installs them.

Dockerfile
raydouglass: Nope, `1000` is fine. I probably just copied-pasted this from somewhere.

ajschmidt8 (commented):
I think this is the last of my comments

.github/workflows/test-notebooks.yml
```diff
@@ -0,0 +1,98 @@
...
+        run: /home/rapids/test_notebooks.py -i /home/rapids/notebooks -o /home/rapids/notebooks_output
+      - name: Install awscli
+        if: '!cancelled()'
+        run: rapids-mamba-retry install -n base awscli
+      - uses: aws-actions/configure-aws-credentials@v1-node16
+        if: '!cancelled()'
+        with:
+          role-to-assume: ${{ vars.AWS_ROLE_ARN }}
+          aws-region: ${{ vars.AWS_REGION }}
+          role-duration-seconds: 43200 # 12h
```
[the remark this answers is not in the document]
raydouglass: Agreed

CONTRIBUTING.md
```diff
@@ -1,86 +1,19 @@
...
-```
-
-If the new repo does not have a `build.sh` file in it's root directory and thus requires special instructions to build, make sure to add an additional `elif` block to the build loop in [Devel.dockerfile.j2](templates/Devel.dockerfile.j2).
-
-## Recommended VS Code Plugin
-
-Users using VS Code as their editor can install the [Better Jinja](https://marketplace.visualstudio.com/items?itemName=samuelcolvin.jinjahtml) plugin to enable Jinja syntax highlighting for `*.dockerfile.j2` files.
+- `CUDA_VER` - Version of CUDA to use. Should be `major.minor.patch`
+- `PYTHON_VER` - Version of Python to use. Should be `major.minor`
+- `LINUX_VER` - Version of Linux to use. Should be one of the options in [`matrix.yaml`](matrix.yaml)
```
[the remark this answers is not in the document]
raydouglass: I've updated a few places for this issue now.

Dockerfile
```diff
@@ -0,0 +1,88 @@
+# syntax=docker/dockerfile:1
+
+ARG CUDA_VER=11.8.0
+ARG PYTHON_VER=3.10
+ARG LINUX_VER=ubuntu22.04
+
+ARG RAPIDS_VER=23.08
+ARG DASK_SQL_VER=2023.6.0
+
+ARG BASE_FROM_IMAGE=rapidsai/mambaforge-cuda
```
ajschmidt8, line 10: does this need to be an `ARG`?

as far as I can tell it's only used in one spot and its value will never change.

raydouglass: No, it doesn't. This was from when I was concurrently updating `mambaforge-cuda` and `docker` to support the `conda` group. Just made it easier to swap the parent image to PR tag on `mambaforge-cuda`.

But I'll remove it.

CONTRIBUTING.md
```diff
@@ -1,86 +1,18 @@
-# RAPIDS Images
+# Contributing to `docker`
```
jakirkham, line 1: Maybe we should indicate it is a RAPIDS project?

```suggestion
# Contributing to `rapidsai/docker`
```

README.md
```diff
@@ -1,75 +1,50 @@
-# <div align="left"><img src="https://rapids.ai/assets/images/RAPIDS-logo-white.svg" width="90px"/>&nbsp;docker
+# `docker`
```
jakirkham, line 1:
```suggestion
# `rapidsai/docker`
```

context/condarc
```diff
@@ -0,0 +1,7 @@
+auto_update_conda: False
+channels:
+  - rapidsai-nightly
+  - dask/label/dev
+  - pytorch
+  - conda-forge
+  - nvidia
```
jakirkham, line 7: Minor style nit to add final newline at EOF

```suggestion
  - nvidia
  
```

jakirkham (commented):
Thanks Ray! 🙏

Couple minor suggestions. Really nice to see all the improvements here

Dockerfile
vyasr: Gotcha. Yeah I'm familiar with the possibility but didn't know that was an intentional design choice here.

context/notebooks.sh
vyasr: I see, that makes sense. Thanks for the explanation.

PR #562, opened 2023-07-27
Remove repos that should not be notebook tested

raydouglass:
Internal discussions resulted that just `cudf`, `cuml`, `cugraph`, & `cuspatial` notebooks should be tested.

We might exclude more notebooks and will iterate on this PR.

### DIFF
--- a/context/notebooks.sh
+++ b/context/notebooks.sh
@@ -8,7 +8,7 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks Ray! 🙏

Is this worth noting in the README or description of these images as well?

raydouglass:
Even though CI is not passing, this PR reduces the number of notebooks being tested, so it will help unblock CI tests in #559

PR #565, opened 2023-08-02
Add input to enable testing

raydouglass:
Refactoring how tests are triggered to use an input. This will allow a calling workflow in `rapidsai/actions` to execute the notebook tests on a nightly basis.

### DIFF
--- a/matrix-test.yaml
+++ b/matrix-test.yaml
@@ -2,7 +2,7 @@ pull-request:
   - { CUDA_VER: '11.2.2', ARCH: 'amd64', PYTHON_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
   - { CUDA_VER: '12.0.1', ARCH: 'amd64', PYTHON_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
   - { CUDA_VER: '11.8.0', ARCH: 'arm64', PYTHON_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
-nightly:
+branch:
   - { CUDA_VER: '11.2.2', ARCH: 'amd64', PYTHON_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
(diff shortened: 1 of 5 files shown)


matrix-test.yaml
```diff
@@ -2,7 +2,7 @@ pull-request:
   - { CUDA_VER: '11.2.2', ARCH: 'amd64', PYTHON_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
   - { CUDA_VER: '12.0.1', ARCH: 'amd64', PYTHON_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
   - { CUDA_VER: '11.8.0', ARCH: 'arm64', PYTHON_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
-nightly:
+branch:
```
ajschmidt8, line 5: what was the reason for switching this key from `nightly` to `branch`?

ajschmidt8: oh nevermind. I see.

it's used as the key here: https://github.com/raydouglass/docker/blob/b7542b9d46badcb3dec468f0b1c09d11c54474c5/.github/workflows/build-test-publish-images.yml#L96


Will this get switch backed to `nightly` after that build and test piplines are split up as mentioned in https://github.com/rapidsai/actions/pull/29?

ajschmidt8 (approved):

matrix-test.yaml
raydouglass: All of the calling workflows use `branch` for the `build_type` and that's the key that is looked up in this file.

raydouglass: Oof, posted before refreshing.

> Will this get switched back to nightly

Yes, I think so.

raydouglass: Also, filed an issue: https://github.com/rapidsai/docker/issues/566

PR #584, opened 2023-08-22
Ensure curl is installed before `gha-tools`

raydouglass:
Not 100% sure why, but [starting recently](https://github.com/rapidsai/docker/actions/runs/5923767292/job/16060805130#step:4:8), `curl` is no longer available in the `notebooks` image during tests.

This PR ensures `curl` & `git` are installed before installing `gha-tools` and running tests. Using `--freeze-installed` ensures the tests run on the same environment that is published with the exception of `curl`, `git`, and `perl` (dependency of `git`) being present, but those packages should not impact the tests.


raydouglass:
The test failures are now on the actual notebooks instead of during setup.

AyodeAwe (approved):

ajschmidt8 (approved):

PR #588, opened 2023-09-07
Add NotebookApp.base_url to jupyter-lab command

### LINKED ISSUE #587: Notebook image no longer works with Kubeflow?
I was using the image at rapidsai/rapidsai:cuda11.8-runtime-ubuntu22.04-py3.10 to run notebooks on a kubeflow server, but after a month or two I had added too many packages and wanted to rebuild it from scratch to fix it a bit.

I noticed that the rapidsai/rapidsai images were deprecated, so I tried using rapidsai/notebooks:23.08-cuda11.8-py3.10 instead. I simply replaced the old image name with the new one.

Unfortunately the notebook server no longer works, giving me the error: Sorry, /notebook/martin/rapids-clustering-only-newimage is not a valid page - ie it's not finding the jupyter server. What do I need to change?

raydouglass:
Closes #587

Adds `--NotebookApp.base_url="${NB_PREFIX:-/}"` to the `jupyter-lab` `CMD` in the `notebooks` image. Since this relies on an environment variable, the `CMD` has to be switched to invoke a shell. See: https://docs.docker.com/engine/reference/builder/#cmd

I think this was missed during the overhaul.

xref: #426

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -84,4 +84,4 @@ EXPOSE 8888
 
 ENTRYPOINT ["/home/rapids/entrypoint.sh"]
 
-CMD ["jupyter-lab", "--notebook-dir=/home/rapids/notebooks", "--ip=0.0.0.0", "--no-browser", "--NotebookApp.token=''", "--NotebookApp.allow_origin='*'"]
+CMD [ "sh", "-c", "jupyter-lab --notebook-dir=/home/rapids/notebooks --ip=0.0.0.0 --no-browser --NotebookApp.token='' --NotebookApp.allow_origin='*' --NotebookApp.base_url=\"${NB_PREFIX:-/}\"" ]


ajschmidt8 (approved):
I see the extra `--NotebookApp.base_url` flag was added.

Why was `"sh", "-c"` added also?

raydouglass:
> Why was `"sh", "-c"` added also?

Since it now relies on an environment variable substitution, the `CMD` has to be switched to invoke a shell. See: https://docs.docker.com/engine/reference/builder/#cmd

ajschmidt8:
> > Why was `"sh", "-c"` added also?
> 
> Since it now relies on an environment variable substitution, the `CMD` has to be switched to invoke a shell. See: https://docs.docker.com/engine/reference/builder/#cmd

Oh whoops. Okay, I see that in your PR body now

jacobtomlinson (approved):
LGTM Thanks @raydouglass

PR #591, opened 2023-10-02
Backporting #590

raydouglass:
We need to do a rebuild of `23.08` to fix CVE-2023-4863.

This requires using the new `rapidsai/ci-conda` images, so just backporting the commit that updated the images.

### DIFF
--- a/.github/workflows/build-test-publish-images.yml
+++ b/.github/workflows/build-test-publish-images.yml
@@ -34,7 +34,7 @@ jobs:
   compute-matrix:
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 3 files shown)


ajschmidt8 (approved):

raydouglass:
The builds are failing because `cucim` (and others) for `23.08` were removed from the `rapidsai-nightly` channel. I ran a local build using the `rapidsai` channel.

However, I did a local build w/ `rapidsai` channel and `libwebp` `1.3.1` was picked up instead of `1.3.2`, so this needs more investigation.

raydouglass:
We opted not to backport the fix.

PR #603, opened 2023-10-16
Fix conda prefix and plotting arguments for `raft-ann-bench` images

dantegd:


raft-ann-bench/gpu/Dockerfile
```diff
@@ -14,24 +14,24 @@ COPY condarc /opt/conda/.condarc
 
 # Create a data folder accessible by any user so mounted volumes under it can be accessible
 # when the user passes their uid to docker run with -u $(id -u)
-RUN mkdir /data && chmod 777 /data
+# Also add the conda_prefix config to the global bashrc file so that all users have it correctly configured.
+RUN mkdir /data && chmod 777 /data \
+    &&  echo "export CONDA_PREFIX=/opt/conda" >> /etc/bash.bashrc
```
raydouglass, line 19: I don't think this is needed because `rapidsai/mambaforge-cuda` will automatically activate the `base` env.

raft-ann-bench/cpu/Dockerfile
```diff
@@ -11,12 +11,15 @@ COPY condarc /opt/conda/.condarc
...
 # when the user passes their uid to docker run with -u $(id -u)
-RUN mkdir /data && chmod 777 /data
-
-# temporarily downgrade conda from 23.9.0 due to https://github.com/mamba-org/mamba/issues/2882
-# after the mamba update step
-RUN mamba update --all -y -n base \
-    && conda install -y -n base conda==23.7.4 \
+# Also add the conda_prefix config to the global bashrc file so that all users have it correctly configured.
+RUN mkdir /data && chmod 777 /data \
+    &&  echo "export CONDA_PREFIX=/opt/conda" >> /etc/bash.bashrc
```
raydouglass, line 16: I think doing this https://github.com/rapidsai/mambaforge-cuda/blob/main/Dockerfile#L34-L38 would be better since it would activate the `base` env automatically.

raft-ann-bench/gpu/Dockerfile
dantegd: I thought so at first but it does for already created users, lik the rapids user environment, it doesn't for the user passed by uid as we found out during testing, which makes this necessary.

raft-ann-bench/cpu/Dockerfile
dantegd: The problem with using the skel file is that that is used for copying files to the home directory of a user at creation time. Since we don't know the user beforehand and it's not created, this wouldn't work. The bash.bashrc file is loaded for all users, even the one passed with -u uid by the user, so that's why need to use this one instead.

raydouglass: Got it. Then you probably want to do this then right (for both cpu & cpu dockerfiles)?
```suggestion
    &&  ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh; \
    &&  echo ". /opt/conda/etc/profile.d/conda.sh; conda activate base" >> /etc/bash.bashrc
```

raydouglass (approved):

PR #609, opened 2023-11-06
Add workflow to publish NGC images

AyodeAwe:
PR adds a workflow to copy `rapidsai/base`, and `rapidsai/notebooks` images to NGC.


AyodeAwe:
I've added my NGC docker secrets and they're used by the workflow to access nvcr.io. I'm not sure though if I'm configured to have push access. I have tested locally and it doesn't look like I have push access to the `rapidsai/notebooks` and `rapidsai/base` repositories.

cc: @raydouglass

.github/workflows/release-to-ngc.yml
```diff
@@ -0,0 +1,41 @@
...
+          password: ${{ secrets.GPUCIBOT_DOCKERHUB_TOKEN }}
+
+    - name: Login to NGC
+      uses: docker/login-action@v3
+      with:
+        registry: nvcr.io
+        username: ${{ secrets.NGC_DOCKER_USER }}
+        password: ${{ secrets.NGC_DOCKER_PASSWORD }}
+
+    - name: Run With DockerHub Token
```
ajschmidt8, line 27: Is this step name accurate? I don't see any token being used

.github/workflows/release-to-ngc.yml
```diff
@@ -0,0 +1,41 @@
...
+      run: |
+        #!/usr/bin/env bash
+        set -e
+        
+        for CUDA in 11.2 11.8 12.0; do
+            for PY in 3.9 3.10; do
```
ajschmidt8, line 33: We should figure out a way to reuse the values in [matrix.yaml](https://github.com/rapidsai/docker/blob/branch-23.12/matrix.yaml) so that we're not redeclaring those values. Otherwise this will become a pain to maintain.

It might be worth exploring converting this into a matrix job.

.github/workflows/release-to-ngc.yml
```diff
@@ -0,0 +1,59 @@
...
+          #!/bin/bash
+          set -e
+
+          CUDA_VER=${{ matrix.CUDA_VER }}
+          CUDA_VER_SHORT=${CUDA_VER%.*}
+          PYTHON_VER=${{ matrix.PYTHON_VER }}
+
+          for type in base notebooks; do
+            source="rapidsai/$type:24.02-cuda$CUDA_VER_SHORT-py$PYTHON_VER"
+            target="nvcr.io/nvidia/rapidsai/$type:23.08-cuda$CUDA_VER_SHORT-py$PYTHON_VER"
```
raydouglass, line 56: The `23.08` should not be hardcoded

raydouglass (changes requested):

.github/workflows/release-to-ngc.yml
raydouglass: And the `24.02` on the line above

PR #622, opened 2023-12-12
Drop CUDA 11.2 from Docker matrix.

bdice:
As communicated in RSN 35, this removes CUDA 11.2 from the Docker support matrix.

https://docs.rapids.ai/notices/rsn0035/

### DIFF
--- a/matrix-test.yaml
+++ b/matrix-test.yaml
@@ -1,14 +1,11 @@
 # CUDA_VER is `<major>.<minor>` (e.g. `11.2`)
 
 pull-request:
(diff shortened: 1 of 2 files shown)


matrix-test.yaml
bdice, line 1: I tried to keep a mix of Python and driver versions while dropping CUDA 11.2.

raydouglass (approved):

bdice:
@raydouglass It looks like some cugraph notebooks are failing. These are known failures, I believe we've seen them in nightlies for some time. Should we admin-merge this PR?

PR #624, opened 2024-01-31
Pin cuda-python

raydouglass:
Fixes attempted in https://github.com/rapidsai/integration/pull/695 & https://github.com/rapidsai/integration/pull/697 don't play well with CEC at large, but this repo can pin `cuda-python` more restrictively.

This PR pins `cuda-python` to the CUDA `major.minor.*` version.

See also https://github.com/conda-forge/cuda-python-feedstock/issues/66

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,7 @@ mamba install -y -n base \
     "rapids=${RAPIDS_VER}.*" \
     "dask-sql=${DASK_SQL_VER%.*}.*" \
     "python=${PYTHON_VER}.*" \
+    # cuda-python pin required for CUDA 12. See https://github.com/conda-forge/cuda-python-feedstock/issues/66
     "cuda-python=${CUDA_VER%.*}.*" \
     "cuda-version=${CUDA_VER%.*}.*" \
     ipython
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,12 +45,12 @@ WORKDIR /home/rapids
 
 COPY condarc /opt/conda/.condarc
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -50,6 +50,7 @@ mamba install -y -n base \
     "rapids=${RAPIDS_VER}.*" \
     "dask-sql=${DASK_SQL_VER%.*}.*" \
     "python=${PYTHON_VER}.*" \
+    "cuda-python=${CUDA_VER%.*}.*" \
```
bdice, line 53: This should be fine, but I'd like for us to add a comment somewhere that indicates this is necessary for CUDA 12 and links to https://github.com/conda-forge/cuda-python-feedstock/issues/66.

bdice (approved):
Add a comment, then this can be merged, I think.

AyodeAwe (approved):

Dockerfile
```diff
@@ -50,6 +50,8 @@ mamba install -y -n base \
     "rapids=${RAPIDS_VER}.*" \
     "dask-sql=${DASK_SQL_VER%.*}.*" \
     "python=${PYTHON_VER}.*" \
+    # cuda-python pin required for CUDA 12. See https://github.com/conda-forge/cuda-python-feedstock/issues/66
```
bdice, line 53: Caution: I don't think you can put a comment in the middle of a bash line continuation `\`. That might be why CI is failing (I didn't look at the logs).

raydouglass: Yeah, I thought it worked because I ran it locally but not long enough for it to fail... I pushed a fix.

PR #632, opened 2024-02-13
fix cugraph notebooks

AyodeAwe:
Certain cugraph notebooks are excluded from testing as is implemented [here](https://github.com/rapidsai/cugraph/issues/2622). This PR updates this repository to reflect that requirement.

### DIFF
--- a/context/test_notebooks.py
+++ b/context/test_notebooks.py
@@ -44,7 +44,7 @@ def get_notebooks(directory: str) -> Iterable[str]:
                     any(match_ignored_file in file for match_ignored_file in ignored_filenames)
                 )
                 and "MNMG" not in file.upper()
-                and not (any(file in ignored_notebook for ignored_notebook in ignored_notebooks))
+                and not any(file in ignored_notebook for ignored_notebook in ignored_notebooks)
             ):
                 path = os.path.join(root.replace(directory, ''), file)
                 if path.startswith('/'):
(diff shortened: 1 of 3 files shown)


context/test_notebooks.py
```diff
@@ -43,7 +44,7 @@ def get_notebooks(directory: str) -> Iterable[str]:
                     any(match_ignored_file in file for match_ignored_file in ignored_filenames)
                 )
                 and "MNMG" not in file.upper()
-                and file not in ignored_notebooks
+                and not (any(file in ignored_notebook for ignored_notebook in ignored_notebooks))
```
bdice, line 47: Typically Python isn’t written with parentheses around the not operator.
```suggestion
                and not any(file in ignored_notebook for ignored_notebook in ignored_notebooks)
```

context/test_notebooks.py
```diff
@@ -43,7 +44,7 @@ def get_notebooks(directory: str) -> Iterable[str]:
                     any(match_ignored_file in file for match_ignored_file in ignored_filenames)
                 )
                 and "MNMG" not in file.upper()
-                and file not in ignored_notebooks
+                and not any(file in ignored_notebook for ignored_notebook in ignored_notebooks)
```
jakirkham, line 47: Would be tempted to rewrite this `if` with `set` operations above on `files`. That would simplify the code improving readability and cutting down on unneeded iteration

bdice:
Do we think these errors will be fixed by https://github.com/conda-forge/ipython-feedstock/pull/210? Is there more work needed after that PR?

raydouglass (approved):

PR #634, opened 2024-02-20
Add CUDA 12.2

jakirkham:
Add CUDA 12.2 images to the build matrix

References:

* https://github.com/rapidsai/build-planning/issues/6
* https://github.com/rapidsai/build-planning/issues/7
* https://github.com/rapidsai/build-planning/issues/8

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -142,7 +142,7 @@ jobs:
             RAPIDS_VER=${{ inputs.RAPIDS_VER }}
           tags: ${{ inputs.RAFT_ANN_BENCH_DATASETS_TAG }}-${{ matrix.ARCH }}
       - name: Build RAFT ANN Benchmarks CPU image
-        if: inputs.CUDA_VER == '12.0.1' # we don't need to build CPU packages for different CUDA versions.
+        if: inputs.CUDA_VER == '12.0.1' || inputs.CUDA_VER == '12.2.2' # we don't need to build CPU packages for different CUDA versions.
         uses: docker/build-push-action@v4
         with:
           context: context
--- a/matrix-test.yaml
+++ b/matrix-test.yaml
@@ -4,8 +4,12 @@ pull-request:
   - { CUDA_VER: '11.8', ARCH: 'amd64', PYTHON_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
   - { CUDA_VER: '12.0', ARCH: 'amd64', PYTHON_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
   - { CUDA_VER: '12.0', ARCH: 'arm64', PYTHON_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
+  - { CUDA_VER: '12.2', ARCH: 'amd64', PYTHON_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
+  - { CUDA_VER: '12.2', ARCH: 'arm64', PYTHON_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
 branch:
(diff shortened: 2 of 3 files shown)


bdice (approved):

jakirkham:
Seeing 2 cuGraph notebook failures happening for any CUDA and Python version. Looks like some changes may have occurred to where test data should be fetched. Asking about these offline

matrix-test.yaml
```diff
@@ -4,8 +4,12 @@ pull-request:
   - { CUDA_VER: '11.8', ARCH: 'amd64', PYTHON_VER: '3.9', GPU: 'v100', DRIVER: 'earliest' }
   - { CUDA_VER: '12.0', ARCH: 'amd64', PYTHON_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
   - { CUDA_VER: '12.0', ARCH: 'arm64', PYTHON_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
+  - { CUDA_VER: '12.2', ARCH: 'amd64', PYTHON_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
+  - { CUDA_VER: '12.2', ARCH: 'arm64', PYTHON_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
```
raydouglass, line 8: I think we should keep the PR matrix small, so just update one of the 12.0 to 12.2 basically.

```suggestion
  - { CUDA_VER: '12.0', ARCH: 'arm64', PYTHON_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
  - { CUDA_VER: '12.2', ARCH: 'amd64', PYTHON_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
```

raydouglass:
> Seeing 2 cuGraph notebook failures happening for any CUDA and Python version

https://github.com/rapidsai/docker/pull/632 might resolve these failures

matrix-test.yaml
bdice: To keep the PR matrix small, I would modify the last line here (amd64 CUDA 12.2) to use Python 3.11 when we're ready for that. Maybe add Python 3.11 for one of the ARM jobs too.

jakirkham: Thinking about this more, the CUDA 12.2 ARM images have more utility than the CUDA 12.0 ARM images. For example cuFile is supported in CUDA 12.2+ (though not earlier). So what if we flipped these?

```suggestion
  - { CUDA_VER: '12.0', ARCH: 'amd64', PYTHON_VER: '3.10', GPU: 'v100', DRIVER: 'latest' }
  - { CUDA_VER: '12.2', ARCH: 'arm64', PYTHON_VER: '3.10', GPU: 'a100', DRIVER: 'latest' }
```

jakirkham: Committing the last change. Happy to discuss further on the updated diff

.github/workflows/build-image.yml
```diff
@@ -142,7 +142,7 @@ jobs:
             RAPIDS_VER=${{ inputs.RAPIDS_VER }}
           tags: ${{ inputs.RAFT_ANN_BENCH_DATASETS_TAG }}-${{ matrix.ARCH }}
       - name: Build RAFT ANN Benchmarks CPU image
-        if: inputs.CUDA_VER == '12.0.1' # we don't need to build CPU packages for different CUDA versions.
+        if: inputs.CUDA_VER == '12.0.1' || inputs.CUDA_VER == '12.2.2' # we don't need to build CPU packages for different CUDA versions.
```
raydouglass, line 145: Only need to build the cpu once, so I think it should use the latest CUDA to key
```suggestion
        if: inputs.CUDA_VER == '12.2.2' # we don't need to build CPU packages for different CUDA versions.
```

raydouglass (approved):

PR #639, opened 2024-03-12
Add labels for interop with AI Workbench

raydouglass:
Adds a list of NVIDIA AI Workbench required labels to the `notebooks` images allowing it to work there with further tweaks.

This required a small change to the compute matrix to separate the linux distro & version.


Dockerfile
```diff
@@ -105,3 +111,32 @@ EXPOSE 8888
...
+LABEL com.nvidia.workbench.cuda-version="$CUDA_VER"
+LABEL com.nvidia.workbench.description="RAPIDS with CUDA ${CUDA_VER}"
+LABEL com.nvidia.workbench.entrypoint-script="/home/rapids/entrypoint.sh"
+LABEL com.nvidia.workbench.labels="cuda${CUDA_VER}"
+LABEL com.nvidia.workbench.name="RAPIDS with CUDA ${CUDA_VER}"
+LABEL com.nvidia.workbench.os="linux"
+LABEL com.nvidia.workbench.os-distro="$LINUX_DISTRO"
+LABEL com.nvidia.workbench.os-distro-release="$LINUX_DISTRO_VER"
+LABEL com.nvidia.workbench.package-manager.apt.binary="/usr/bin/apt"
+LABEL com.nvidia.workbench.package-manager.apt.installed-packages="curl git git-lfs vim"
```
dkleissa, line 133: these packages are not in this container. Either remove them from this label, or install.

Dockerfile
```diff
@@ -105,3 +111,32 @@ EXPOSE 8888
...
+LABEL com.nvidia.workbench.package-manager.pip.binary="/opt/conda/bin/pip"
+LABEL com.nvidia.workbench.package-manager.pip.installed-packages=""
```
dkleissa, line 135: while optional, if you list your top-level installed packages here they will show up in the UI so users know what packages and what versions are installed.

dkleissa (commented):
You should add the `image-version` label to support auto-updates in the future. Since users can have whatever arbitrary tagging schema, the `image-version` label provides a way to communicate versions of the image as they are published. It is a simple semantic version string that is then used to know if a newer version of the image is available. If so, a button to "update" is presented to the user.

```dockerfile
LABEL com.nvidia.workbench.image-version="1.0.0"
```
When building & publishing a new image, you must increment this value by whatever makes sense to you. As long as the increasing versions will sort as expected it should work for you.


Suggest adding this line to prevent the notices pop up every time the user starts jupyter:

```dockerfile
# Disable the announcement banner
RUN /opt/conda/bin/jupyter labextension disable "@jyupyterlab/apputils-extension:announcements"
```


This image will be published for both arm64 and amd64, correct?

AyodeAwe (approved):

Dockerfile
```diff
@@ -95,6 +101,9 @@ conda clean -afy
 pip cache purge
 EOF
 
+# Disable the JupyterLab announcements
+RUN /opt/conda/bin/jupyter labextension disable "@jyupyterlab/apputils-extension:announcements"
```
bdice, line 105: Is this a typo?
```suggestion
RUN /opt/conda/bin/jupyter labextension disable "@jupyterlab/apputils-extension:announcements"
```

raydouglass: Oof, copy-paste error. And bit disappointed this did not fail the build.

PR #648, opened 2024-04-02
Add retry to mamba installs

raydouglass:
The docker builds are pretty consistently encountering `Download error (28) Timeout was reached` errors. It almost always occurs on a large RAPIDS package but that makes sense since they take the longest to download. I was able to reproduce locally, so I don't think it's a problem with our runners. I'm going to reach out to Anaconda separately.

This PR sets `remote_max_retries` so that `mamba` will retry which should help with the issue.

This also sets `MAMBA_NO_LOW_SPEED_LIMIT=1` which prevents `mamba` from erroring out if a download is too slow. Originally, I wasn't sure abut using this, but GHA times out on its own, so it should be safe.

Examples of the errors:
- https://github.com/rapidsai/docker/actions/runs/8466203792/job/23194501903#step:9:651
- https://github.com/rapidsai/docker/actions/runs/8506470533/job/23296743943#step:9:668
- https://github.com/rapidsai/docker/actions/runs/8520794283/job/23345886575#step:9:219 (dask repodata file that's only ~268kb)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,7 +52,7 @@ WORKDIR /home/rapids
 COPY condarc /opt/conda/.condarc
 
 RUN <<EOF
-mamba install -y -n base \
+MAMBA_NO_LOW_SPEED_LIMIT=1 mamba install -y -n base \
     "rapids=${RAPIDS_VER}.*" \
     "dask-sql=${DASK_SQL_VER%.*}.*" \
     "python=${PYTHON_VER}.*" \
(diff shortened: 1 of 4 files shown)


raydouglass:
The changes so far "fixed" the problems in the base & notebook images. But now getting stranger errors.

```
#11 306.2 error    libmamba ZSTD decompression error: Unknown frame descriptor
#11 306.2 Download error (23) Failed writing received data to disk/application [https://conda.anaconda.org/dask/label/dev/noarch/repodata.json.zst]
#11 306.2 Failure writing output to destination, passed 689 returned 690
```

PR #649, opened 2024-04-02
Include Conda debug info before install

jakirkham:
This is common information asked for by Conda issue templates (like [`conda/infrastructure` bug reports]( https://github.com/conda/infrastructure/issues/new?assignees=&labels=type%3A%3Abug&projects=&template=0_bug.yml )). Query it during the build so we have this information in the logs when sharing them with others to debug issues.


jakirkham:
We may want to add more lines to this in the future

Also it is possible we want this in more places. Raising tracking issue: https://github.com/rapidsai/build-planning/issues/34

raydouglass (approved):

msarahan (approved):
This is a good idea, but might it be better to move it to the existing heredoc below, instead of adding another image layer?

jakirkham:
Interesting would have thought we used squash, but it appears that is not possible with the action we use: https://github.com/docker/build-push-action/issues/56

Dockerfile
```diff
@@ -53,6 +53,12 @@ WORKDIR /home/rapids
 
 COPY condarc /opt/conda/.condarc
 
+RUN <<EOF
+conda info
+conda config --show-sources
+conda list --show-channel-urls
+EOF
+
 RUN <<EOF
```
jakirkham, line 62: Meaning something like this?

```suggestion
```

jakirkham:
This is now in one `RUN` command / layer

Guessing this is ready to go?

msarahan (approved):

jakirkham:
Thanks Michael! 🙏

PR #682, opened 2024-06-07
Adopt packaging codeowners

bdice:
This PR overhauls how `ops-codeowners` reviews are handled.

`ops-codeowners` is replaced by `ci-codeowners` & `packaging-codeowners`. The coverage of files is expanded as well.

Additionally, the process will change: reviews will be assigned to a member of the teams instead of a manual request to `ops-codeowners`.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
-#ops code owners
-*       @rapidsai/ops-codeowners
+# packaging code owners
+* @rapidsai/packaging-codeowners


jameslamb:
I support this. The images published to Docker Hub from this repo with tags like `:24.06a-cuda12.0-py3.11` are "packages" with the same visibility to downstream users of RAPIDS at the nightly conda packages and wheels, so I think having a wider group of codeowners be able to approve (the same group that approves packaging changes to those conda packages and wheels) makes sense.

raydouglass (approved):

PR #684, opened 2024-06-25
Remove dask-sql

raydouglass:
Since `dask-sql` does not have a lead maintainer, we are dropping it from our containers.

User who wish to continue using `dask-sql` in the containers can install it manually or use the `EXTRA_CONDA_PACKAGES` environment variable to install it on startup.

See: https://github.com/dask-contrib/dask-sql/issues/1344


raydouglass:
The test failures are unrelated since I see the same ones in the nightly builds and `cuspatial` doesn't use `dask-sql`. I've asked the `cuspatial` team to take a look.

.github/workflows/build-image.yml
```diff
@@ -70,8 +66,8 @@ jobs:
           GIT_DESCRIBE_TAG="$(git describe --tags --abbrev=0)"
           GIT_DESCRIBE_TAG="${GIT_DESCRIBE_TAG:1}" # remove leading 'v'
           if [[ ! $GIT_DESCRIBE_TAG =~ [a-z] ]]; then
-            rapids-logger "Most recent tag is for release, adding `rapidsai` channel and removing `rapidsai-nightly` and `dask/label/dev` channels."
-            sed -i 's|rapidsai-nightly|rapidsai|;\|dask/label/dev|d' context/condarc
+            rapids-logger "Most recent tag is for release, adding `rapidsai` channel and removing `rapidsai-nightly` channel."
```
bdice, line 69: Was dask-sql the only package we needed this channel for? Are packages like dask, distributed either (1) not installed or (2) available from other channels?

raydouglass: Oh this is a good point. The `dask/label/dev` channel does have dask/distributed alphas which are [used in builds](https://github.com/rapidsai/docker/actions/runs/9676448229/job/26696030096#step:9:626). I thought we added this channel just for `dask-sql`, but looking through git history, that doesn't seem to be the case, so I will revert this change.

context/condarc
```diff
@@ -1,7 +1,6 @@
 auto_update_conda: False
 channels:
   - rapidsai-nightly
-  - dask/label/dev
```
bdice, line 4: @raydouglass Is this supposed to be reverted, too?

raydouglass: Yes, good catch!

bdice (approved):

PR #689, opened 2024-07-16
add CUDA 12.5 images

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/73

Proposes the following:

* adding CUDA 12.5 images
* PR builds:
    * +1 test job covering `(cuda=12.5, arch=arm64)`
* branch builds:
    * +1 test job covering `(cuda=12.5, arch=x86_64)`
    * -1  test job covering `(cuda=12.2, arch=arm64)`

context: https://github.com/rapidsai/build-planning/issues/73#issuecomment-2231443095

## Notes for Reviewers

Per offline discussion with @raydouglass , this would be accompanied by a deprecation notice in the RAPIDS 24.08 release stating that the CUDA 12.2 images will be removed in some future release (future release not yet determined).

PR #690, opened 2024-07-17
fix return-before-assignment bug in notebook test script, add pre-commit, drop cuspatial notebooks

jameslamb:
CI has been failing here for a few weeks, as a result of some failing `cuspatial` notebooks.

As described in https://github.com/rapidsai/cuspatial/issues/1406, I observed those notebooks failing like this:

```text
cannot access local variable 'execution_time' where it is not associated with a value
```

Turns out that that was coming from the `test_notebooks.py` script in this repo. It calls `return execution_time` unconditionally, even though it's possible for that variable to not exist.

This fixes that, and proposes adding a `pre-commit` configuration and corresponding CI job, to help catch such things earlier in the future.

This also proposes skipping some `cuspatial` notebooks, to get CI working. I'd hoped to address that in https://github.com/rapidsai/cuspatial/pull/1407, but most `cuspatial` maintainers are busy or unavailable this week. I think we should get CI working here again and deal with the `cuspatial` notebooks as a separate concern.

## Notes for Reviewers

Any other changes you see in files here are the result of running `pre-commit run --all-files` (with some manual fixes applied to fix `ruff` warnings).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -21,7 +21,7 @@ A clear and concise description of what you expected to happen.
  - Method of cuDF install: [conda, Docker, or from source]
    - If method of install is [Docker], provide `docker pull` & `docker run` commands used
  - Please run and attach the output of the `cudf/print_env.sh` script to gather relevant environment details
- 
+
 
 **Additional context**
 Add any other context about the problem here.
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -10,7 +10,20 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+      - name: Run pre-commit
+        run: |
+          pip install pre-commit
+          pre-commit run --all-files
   docker:
+    needs: [checks]
     uses: ./.github/workflows/build-test-publish-images.yml
     with:
       build_type: pull-request
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,12 @@
+---
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.6.0
+    hooks:
+      - id: end-of-file-fixer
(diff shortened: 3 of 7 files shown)


raydouglass (approved):

jakirkham (approved):

jakirkham:
One of the jobs had startup issues. Restarted

jameslamb:
blegh I see what happened.

One of the test jobs is failing like this:

>  Error response from daemon: manifest for rapidsai/staging:docker-notebooks-690-24.08a-cuda12.0-py3.10-amd64 not found: manifest unknown: manifest unknown

([build link](https://github.com/rapidsai/docker/actions/runs/9993739485/job/27630482326?pr=690))

I'm thinking that maybe the corresponding `delete-temp-images` job already ran.

Screenshot 2024-07-18 at 1 00 54 PM

*source:* https://github.com/rapidsai/docker/actions/runs/9993739485?pr=690

Looks like #574 tried to fix that, but I suspect that the `if: always()` results in the behavior "run the deletion even if tests failed".

https://github.com/rapidsai/docker/blob/a1e5d98caf1efa58c5886659ff950cfcd5b7dd7e/.github/workflows/build-test-publish-images.yml#L284-L286

I think it'll take a full re-run to get CI passing here. I just kicked one off here.

ajschmidt8 (approved):
thanks James

jakirkham:
Thanks James! 🙏

Regarding CI, maybe we should write this up as an issue and see if there is a more friendly way to fix these one-off failures. Remember seeing this behavior before

jakirkham:
Oops thought it would wait for CI 😳

At least we know nearly all passed and the one that failed was mostly likely a fluke

Will keep an eye on CI on `branch-24.08`

jameslamb:
No problem, thanks John.

Just noting that the tests don't run on merges to `main`:

* https://github.com/rapidsai/docker/actions/runs/9996638512
* https://github.com/rapidsai/docker/blob/ddb36fd24f8b8333dca0ee22b1412cf1f00d370c/.github/workflows/publish.yml#L25

So we'll have to check the CI of another PR (like #689) to know if this worked.

jakirkham:
Hmm...good point. Seems like we have a few footguns to work on here 😅

Well it does look like most of those passed (except CUDA 12.5 where we still need cuSpatial)

Opened this PR ( https://github.com/rapidsai/docker/pull/692 ) to test CI with no change

jameslamb:
Yep definitely, thanks for helping us uncover them!

PR #688, opened 2024-07-03
resolve warnings about casing in Dockerfiles

jameslamb:
I've been seeing this build warning in logs here:

```text
 2 warnings found (use --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 37)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 70)
```

([example build link](https://github.com/rapidsai/docker/actions/runs/9773984230/job/26981297565#step:9:826))

That comes from BuildKit's build checks (https://docs.docker.com/reference/build-checks/). Details on that specific warning: https://docs.docker.com/reference/build-checks/from-as-casing/.

This fixes it, to remove a source of noise in logs (and because I agree with the suggestion that capitalizing keywords approves readability a bit).

PR #692, opened 2024-07-18
[NO MRG] Test CI

jakirkham:


jakirkham:
One of the jobs had trouble during building finding the `rapids=24.08` package. Likely a network issue as the packages appear to be there and we haven't had issues in other builds. Restarted all jobs to have a clean build

jameslamb:
thanks @jakirkham .

I just remembered I have #688, want to use that to continue testing CI? It isn't adding CUDA 12.5 support so shouldn't be blocked by `cuspatial`, and it at least should eventually be merged (to fix some very minor stylistic warnings).

I restarted CI there a few minutes ago.

jakirkham:
We can certainly do that

Looks like we are down to the final 3 jobs here. So may just wait and see what happens

jameslamb:
Looks like tests did pass here and in https://github.com/rapidsai/docker/pull/688! I'm going to merge https://github.com/rapidsai/docker/pull/688, and I think we can close this.

PR #693, opened 2024-07-18
add pr-builder job

### LINKED ISSUE #691: Configure `pr-builder` for `rapidsai/docker`
Recently noticed that `rapidsai/docker` is not configured to use `pr-builder`. As a result, commenting `/merge` will merge a PR before CI passes ( https://github.com/rapidsai/docker/pull/690#issuecomment-2237210976 ). This differs from other RAPIDS repos

To make sure PRs only merge after CI passes, it would be good to configure `pr-builder` on this repo

jameslamb:
fixes #691

Proposes adding the `pr-builder` workflow, to ensure that all required checks have passed prior to merging (or auto-merging!) here.

This makes auto-merger in this repo behave the way it does across most other repos (https://docs.rapids.ai/resources/auto-merger/).

In exchange, it means all changes will now require a full CI run to be merged. That can take 1-2 hours ([list of recent runs](https://github.com/rapidsai/docker/actions/workflows/pr.yml?query=is%3Asuccess)).

## Manual actions before merging

* [ ] an admin needs to add branch protection(s) blocking merge if the `pr-builder` workflow does not succeed

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -10,6 +10,12 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pr-builder:
+    needs:
+      - checks
+      - docker
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.08
   checks:
     runs-on: ubuntu-latest
     steps:
@@ -29,9 +35,3 @@ jobs:
       build_type: pull-request
       run_tests: true
     secrets: inherit
-  pr-builder:
-    needs:
-      - checks
-      - docker
-    secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.08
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -29,3 +29,9 @@ jobs:
       build_type: pull-request
       run_tests: true
     secrets: inherit
+  pr-builder:
```
raydouglass, line 32: Just a nitpick, but I think all other repos put this job first.

raydouglass (approved):
One nitpick, but approving otherwise.

jameslamb:
Ok sure, no problem.

I'd put `pr-builder` last in the file because it's the last workflow in the graph to run (as it depends on all others), so to me it felt natural to have the configuration flow from top to bottom in the same order that the workflow dependencies flow from left to right.

But I don't feel that strongly about it.

jakirkham:
Thanks James and Ray! 🙏

PR #698, opened 2024-08-04
enable jobs retry & multi-arch builds of succeeding image variants

AyodeAwe:
WIP


.github/workflows/build-test-publish-images.yml
```diff
@@ -191,10 +191,11 @@ jobs:
         py${{ matrix.PYTHON_VER }}-\
         ${{ matrix.ARCH }}"
   build-multiarch-manifest:
-    if: inputs.build_type == 'branch'
```
AyodeAwe, line 194: Will make this `build-multiarch-manifest` job `branch` only after this PR is reviewed.

AyodeAwe:
Discovered a problem where when the build is re-run, all matrices of the downstream jobs (whether previously passing or failing) are re-triggered.

This is an issue because if we make the delete job without all the build/multi-arch jobs passing, when we rerun failed jobs, the previously passing build/multi-arch jobs will now fail (given that in the previous run, the delete job deleted all the temporary images for the passing builds)

The current re-adjustment now is to make the delete job explicitly wait for the test jobs (which depends on the multi-arch jobs) to pass.

AyodeAwe:
PR is continued [here](https://github.com/rapidsai/docker/pull/702) (re-targeted for 24.10).

PR #702, opened 2024-08-07
enable jobs retry & multi-arch builds of succeeding docker image builds

AyodeAwe:
**Approach**:

1. Enable the construction of multi-architecture images even if not all image build matrices are successful. This allows for the publishing of manifests for the images that have been successfully built without waiting for all builds to complete.
    - In this change, we also modify the multi-arch job to only generate a manifest when the necessary images are available; otherwise, we make the job fail.
2. Modify the deletion job to execute only after the publishing and tests jobs succeed. This prevents the need to restart all jobs when a single image build job fails.

**Here are the new behaviors**:
- The `build-multiarch-manifest` no longer depends on the `build` jobs to all pass before publishing multi-arch images to dockerhub. Once the `build` job completes (either succeeds or fails), the multiarch job builds and pushes manifests for the succeeding images.
- The `test` job _still_ uses the images from `build` to run tests.
- The `delete-temp-images` job will only run AFTER `build-multiarch-manifest` and `test` jobs succeed.
- If any jobs fail during the workflow run, we can simply `re-run all failed jobs` now. Once all jobs pass, the delete job will trigger.

### DIFF
--- a/ci/create-multiarch-manifest.sh
+++ b/ci/create-multiarch-manifest.sh
@@ -1,0 +1,78 @@
+# Authenticate and retrieve DockerHub token
+HUB_TOKEN=$(
+curl -s -H "Content-Type: application/json" \
+    -X POST \
+    -d "{\"username\": \"$GPUCIBOT_DOCKERHUB_USER\", \"password\": \"$GPUCIBOT_DOCKERHUB_TOKEN\"}" \
+    https://hub.docker.com/v2/users/login/ | jq -r .token \
+)
+echo "::add-mask::${HUB_TOKEN}"
+
+org="rapidsai"
+
+# Initialize arrays to store source tags for each image
+base_source_tags=()
+notebooks_source_tags=()
+raft_ann_bench_source_tags=()
+raft_ann_bench_datasets_source_tags=()
+raft_ann_bench_cpu_source_tags=()
+
+# Define tag arrays for different images
+base_tag="${BASE_TAG_PREFIX}${RAPIDS_VER}${ALPHA_TAG}-cuda${CUDA_TAG}-py${PYTHON_VER}"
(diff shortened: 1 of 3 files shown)


raydouglass:
I re-targeted for `branch-24.08` because we're having lots of issues with the release builds (https://github.com/rapidsai/docker/actions/runs/10302850714).

raydouglass (approved):

ci/create-multiarch-manifest.sh
```diff
@@ -0,0 +1,78 @@
+# Authenticate and retrieve DockerHub token
```
jameslamb, line 1: I think this would benefit from more strictness.

```bash
#!/bin/bash

set -e -E -u -o pipefail
```

would you consider that?

ci/delete-temp-images.sh
```diff
@@ -0,0 +1,46 @@
+# Authenticate and retrieve DockerHub token
```
jameslamb, line 1: Same here, this would benefit from some strictness so it'll fail faster, and nearer to the source of any issues.

```bash
#!/bin/bash

set -e -E -u -o pipefail
```

jameslamb (approved):
Left some small suggestions, all could be addressed in a follow-up on 24.10.

I have one question:

> *The delete-temp-images job will only run AFTER build-multiarch-manifest and test jobs succeed.*

By itself, this would mean that artifacts from failed builds are left behind on DockerHub forever, wouldn't it? Do we today have some other mechanism that cleans up old intermediate CI artifacts? If not, we probably should do that (can be later work, doesn't have to block this PR). I wouldn't want our ability to publish new images to be stopped by hitting some service quotas from DockerHub, or to be paying indefinitely for things that we don't need.

AyodeAwe:
Answered @jameslamb in [Slack](https://nvidia.slack.com/archives/C01PF84NXCL/p1723128341103179?thread_ts=1723127395.163929&cid=C01PF84NXCL).

PR #706, opened 2024-08-08
incorporate build condition for raft-cpu-image in multi-arch job

AyodeAwe:
Accounts for [this condition](https://github.com/rapidsai/docker/blob/branch-24.08/.github/workflows/build-image.yml#L149) for the multiach-manifest job.

### DIFF
--- a/.github/workflows/build-test-publish-images.yml
+++ b/.github/workflows/build-test-publish-images.yml
@@ -133,6 +133,7 @@ jobs:
       LINUX_VER: ${{ matrix.LINUX_VER }}
       PYTHON_VER: ${{ matrix.PYTHON_VER }}
       RAPIDS_VER: ${{ needs.compute-matrix.outputs.RAPIDS_VER }}
+      BUILD_RAFT_ANN_BENCH_CPU_IMAGE: ${{ matrix.BUILD_RAFT_ANN_BENCH_CPU_IMAGE }}
       BASE_TAG:
         "rapidsai/${{ needs.compute-matrix.outputs.BASE_IMAGE_REPO }}:\
         ${{ needs.compute-matrix.outputs.BASE_TAG_PREFIX }}\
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test-publish-images.yml
```diff
@@ -187,6 +187,7 @@ jobs:
       - name: Create multiarch manifest
         shell: bash
         env:
+          RAFT_ANN_BENCH_CPU_IMAGE_BUILT: ${{ matrix.CUDA_TAG == '12.5.1' }}
```
jameslamb, line 190: Instead of duplicating this `12.5.1` in another place, would you consider using GitHub Actions `outputs`?

I think it should be possible to do something like this example:

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/defining-outputs-for-jobs#using-job-outputs-in-a-matrix-job

Where the `build` job could directly set an output `RAFT_ANN_BENCH_CPU_IMAGE_BUILT` or similar that could then be referenced here.

I think that'd be preferable, because at first glance without the context from the PR description, it's not clear why a specific CUDA version is showing up in this part of the configuration.

jameslamb (changes requested):
Left one suggestion for your consideration.

.github/workflows/build-test-publish-images.yml
AyodeAwe: @jameslamb I've moved the logic to the jq script -- we effectively determine if we should build that image (depending on the CUDA version) right from where we construct the matrix. Thoughts?

ci/compute-matrix.jq
```diff
@@ -14,6 +14,9 @@ def compute_ubuntu_version($x):
 def compute_cuda_tag($x):
   $x + {CUDA_TAG: $x.CUDA_VER | split(".") | [.[0], .[1]] | join(".") };
 
+def compute_build_raft_ann_bench_cpu_image($x):
+  $x + {BUILD_RAFT_ANN_BENCH_CPU_IMAGE: ($x.CUDA_VER == "12.5.1")}; # we don't need to build CPU packages for different CUDA versions.
```
jameslamb, line 18: Thanks for single-sourcing this, I like this approach of handling it here along with the other matrix operations.

Is it possible to go even further and make this condition something like "sort all options by `CUDA_VER` and take the latest CUDA version"?

That way, as `12.5.1` drops out of the matrix in the future, this would continue to work without modifications required.

My `jq` is not very good so I can't recommend the exact syntax to use here, but I know we've done something similar in other CI configs like this:

```yaml
# Package is pure Python and only ever requires one build.
# This selects "ARCH=amd64 + the latest supported Python + CUDA".
matrix_filter: map(select(.ARCH == "amd64")) | max_by([(.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(tonumber))]) | [.]
```

https://github.com/rapidsai/dask-cuda/blob/00c37dc55bee1a34f7d9f6599a89a3f89c15651b/.github/workflows/pr.yaml#L52-L54

jameslamb (commented):
Thanks for this! Now that I see this happening down in a `.jq` script, proposed one more change for your consideration.

raydouglass (approved):

jameslamb (approved):
Thanks! Changes look good to me, hopefully we'll get a successful CI run to verify that the filter worked as expected. Looks like the recent builds mostly failed with network-related issues.

PR #707, opened 2024-08-09
resolve warnings about Dockerfile casing

jameslamb:
Reviewing #706, I noticed the following on the Files tab (https://github.com/rapidsai/docker/pull/706/files):

> FromAsCasing: 'as' and 'FROM' keywords' casing do not match
> More info: https://docs.docker.com/go/dockerfile/rule/from-as-casing/

image

This resolves those warnings. They don't change anything functionally, but I do slightly agree with the style suggestion, and it'd be nice to remove the visual noise of all those warnings in PR reviews.


bdice (approved):
Thanks! I noticed this too!!

PR #696, opened 2024-07-26
Drop CUDA 12.2 Docker images

bdice:
Following https://github.com/rapidsai/docs/pull/526, we can remove CUDA 12.2 from the RAPIDS 24.10 Docker images.

PR #708, opened 2024-08-14
ensure test jobs always run on PRs

jameslamb:
Follow-up to #702 and #693.

Created based on https://github.com/rapidsai/docker/pull/696#issuecomment-2289689619

`test` jobs are not currently running on pull requests here, because they require `build-multiarch-manifest` jobs, which have this condition that causes such jobs to be skipped on PR builds:

https://github.com/rapidsai/docker/blob/1c27d9245fd9d99ee35981b970acaf10961ca45b/.github/workflows/build-test-publish-images.yml#L171-L172

This PR ensures that `test` jobs always run on PRs, and that merging is blocked until they succeed.

### DIFF
--- a/.github/workflows/build-test-publish-images.yml
+++ b/.github/workflows/build-test-publish-images.yml
@@ -38,6 +38,7 @@ jobs:
       - compute-matrix
       - build
       - test
+      - delete-temp-images
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.10
   checks:
@@ -249,7 +250,8 @@ jobs:
         ${{ needs.compute-matrix.outputs.RAPIDS_VER }}\
         ${{ needs.compute-matrix.outputs.ALPHA_TAG }}-\
         cuda${{ matrix.CUDA_VER }}-\
-        py${{ matrix.PYTHON_VER }}"
+        py${{ matrix.PYTHON_VER }}"-\
+        ${{ matrix.ARCH }}
   delete-temp-images:
     if: ${{ !cancelled() && needs.test.result == 'success' }}
     needs: [compute-matrix, build, test]
--- a/.github/workflows/build-test-publish-images.yml
+++ b/.github/workflows/build-test-publish-images.yml
@@ -31,6 +31,13 @@ permissions:
   statuses: none
 
 jobs:
+  pr-builder:
+    needs:
+      - compute-matrix
+      - build
+      - test
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.10
   compute-matrix:
     runs-on: ubuntu-latest
     container:
@@ -208,7 +215,7 @@ jobs:
           ARCHES: ${{ toJSON(matrix.ARCHES) }}
         run: ci/create-multiarch-manifest.sh
   test:
-    needs: [compute-matrix, build, build-multiarch-manifest]
+    needs: [compute-matrix, build]
     if: inputs.run_tests
     strategy:
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.TEST_MATRIX) }}
@@ -231,7 +238,7 @@ jobs:
         py${{ matrix.PYTHON_VER }}"
   delete-temp-images:
     if: ${{ !cancelled() && needs.test.result == 'success' }}
-    needs: [compute-matrix, build-multiarch-manifest, test]
+    needs: [compute-matrix, build, test]
     strategy:
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
       fail-fast: false
--- a/.github/workflows/build-test-publish-images.yml
+++ b/.github/workflows/build-test-publish-images.yml
@@ -250,8 +250,8 @@ jobs:
         ${{ needs.compute-matrix.outputs.RAPIDS_VER }}\
         ${{ needs.compute-matrix.outputs.ALPHA_TAG }}-\
         cuda${{ matrix.CUDA_VER }}-\
-        py${{ matrix.PYTHON_VER }}"-\
-        ${{ matrix.ARCH }}
+        py${{ matrix.PYTHON_VER }}-\
+        ${{ matrix.ARCH }}"
   delete-temp-images:
     if: ${{ !cancelled() && needs.test.result == 'success' }}
     needs: [compute-matrix, build, test]
--- a/.github/workflows/build-test-publish-images.yml
+++ b/.github/workflows/build-test-publish-images.yml
@@ -32,12 +32,26 @@ permissions:
 
 jobs:
   pr-builder:
+    if: ${{ !cancelled() && inputs.build_type == 'pull-request' }}
     needs:
+      - checks
       - compute-matrix
       - build
       - test
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.10
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+      - name: Run pre-commit
+        run: |
+          pip install pre-commit
+          pre-commit run --all-files
   compute-matrix:
     runs-on: ubuntu-latest
     container:
@@ -126,7 +140,7 @@ jobs:
 
           echo "TEST_MATRIX=$(yq -n -o json 'env(TEST_MATRIX)' | jq -c '{include: .}')" | tee --append "${GITHUB_OUTPUT}"
   build:
-    needs: compute-matrix
+    needs: [checks, compute-matrix]
     strategy:
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
       fail-fast: false
--- a/.github/workflows/build-test-publish-images.yml
+++ b/.github/workflows/build-test-publish-images.yml
@@ -37,6 +37,7 @@ jobs:
       - checks
       - compute-matrix
       - build
+      - build-multiarch-manifest
       - test
       - delete-temp-images
     secrets: inherit
@@ -46,9 +47,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.10"
       - name: Run pre-commit
         run: |
           pip install pre-commit
@@ -191,7 +189,6 @@ jobs:
         ${{ needs.compute-matrix.outputs.ALPHA_TAG }}-\
         py${{ matrix.PYTHON_VER }}"
   build-multiarch-manifest:
-    if: ${{ !cancelled() && inputs.build_type == 'branch' }}
     needs: [build, compute-matrix]
     strategy:
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
@@ -253,8 +250,8 @@ jobs:
         py${{ matrix.PYTHON_VER }}-\
         ${{ matrix.ARCH }}"
   delete-temp-images:
-    if: ${{ !cancelled() && needs.test.result == 'success' }}
-    needs: [compute-matrix, build, test]
+    if: ${{ !cancelled() && (needs.test.result == 'success' || needs.test.result == 'skipped') }}
+    needs: [compute-matrix, build, build-multiarch-manifest, test]
     strategy:
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
       fail-fast: false
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -10,26 +10,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  pr-builder:
-    needs:
-      - checks
-      - docker
-    secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.10
-  checks:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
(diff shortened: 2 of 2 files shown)


jameslamb:
It looks to me like the `test` jobs either need to depend on the `build-multiarch-manifest` job, or need to include architecture in their expected image tags.

See this failing build:

```text
/usr/local/bin/docker pull rapidsai/staging:docker-notebooks-708-24.10a-cuda11.8-py3.9
  Error response from daemon: manifest for rapidsai/staging:docker-notebooks-708-24.10a-cuda11.8-py3.9 not found: manifest unknown: manifest unknown
  Warning: Docker pull failed with exit code 1, back off 4.989 seconds before retry.
  /usr/local/bin/docker pull rapidsai/staging:docker-notebooks-708-24.10a-cuda11.8-py3.9
  Error response from daemon: manifest for rapidsai/staging:docker-notebooks-708-24.10a-cuda11.8-py3.9 not found: manifest unknown: manifest unknown
  Warning: Docker pull failed with exit code 1, back off 9.9[16](https://github.com/rapidsai/docker/actions/runs/10393969779/job/28784483770?pr=708#step:3:19) seconds before retry.
  /usr/local/bin/docker pull rapidsai/staging:docker-notebooks-708-24.10a-cuda11.8-py3.9
  Error response from daemon: manifest for rapidsai/staging:docker-notebooks-708-24.10a-cuda11.8-py3.9 not found: manifest unknown: manifest unknown
  Error: Docker pull failed with exit code 1
```

([build link](https://github.com/rapidsai/docker/actions/runs/10393969779/job/28784483770?pr=708))

There are only architecture-specific tags available at https://hub.docker.com/r/rapidsai/staging/tags?page=&page_size=&ordering=&name=docker-notebooks-708-24.10a-cuda11.8-py3.9.

Screenshot 2024-08-14 at 3 43 53 PM

I just pushed https://github.com/rapidsai/docker/pull/708/commits/11d05f7fe95000cd28557d99af0fdb49c9f1f335 adding the architecture to `NOTEBOOK_TAG`. Since in #702 it looks like the `build-multiarch-manifest` job was intentionally skipped for PRs.

jameslamb:
Here's an example where one build failed with a network error, and where I was able to re-run only the failed job (the main design goal of #702):

Screenshot 2024-08-14 at 4 33 31 PM

https://github.com/rapidsai/docker/actions/runs/10394821734/job/28785524564?pr=708

And I can see that the `delete-*` jobs have not run yet, so all the images needed by the `test` jobs should exist.

jameslamb:
Ok I think this is working!

✅ was able to run "re-run failed jobs" on `build` jobs (no need to re-run all of them)
✅ all `test` jobs ran
✅ 0 `test` jobs ran until all `build` jobs passed
✅ 0 `delete-temp-images` job ran until all `test` jobs passed
✅ all `delete-temp-images` ran successfully, and there are 0 images left behind from this PR ([DockerHub search for '-708' tags for 'rapidsai/staging'](https://hub.docker.com/r/rapidsai/staging/tags?page=&page_size=&ordering=&name=-708))

jameslamb:
@raydouglass @AyodeAwe if you agree with the proposal in this PR, could you please change which `pr-builder` is required in the branch protection checks? It should now be `ci / docker / pr-builder / run`.

Screenshot 2024-08-15 at 10 10 13 AM

.github/workflows/pr.yml
```diff
@@ -10,26 +10,7 @@ concurrency:
...
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.10"
-      - name: Run pre-commit
-        run: |
-          pip install pre-commit
-          pre-commit run --all-files
```
jameslamb, line 30: moving this so we can have finer-grained control over what has to pass.

Having this just depend on `docker` means it passed even when the `test` jobs in `docker` were skipped.

AyodeAwe (approved):

.github/workflows/build-test-publish-images.yml
```diff
@@ -31,6 +31,28 @@ permissions:
...
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.10
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
```
bdice, line 51: Might as well delay updating by one more minor version.
```suggestion
          python-version: "3.11"
```

.github/workflows/build-test-publish-images.yml
```diff
@@ -31,6 +31,28 @@ permissions:
   statuses: none
 
 jobs:
+  pr-builder:
+    if: ${{ !cancelled() && inputs.build_type == 'pull-request' }}
+    needs:
+      - checks
+      - compute-matrix
+      - build
```
bdice, line 39: Are we missing `build-multiarch-manifest` here?

.github/workflows/build-test-publish-images.yml
jakirkham: Yeah was wondering if we should do this

Is there value in running this with the oldest Python we support?

.github/workflows/build-test-publish-images.yml
```diff
@@ -228,10 +250,11 @@ jobs:
         ${{ needs.compute-matrix.outputs.RAPIDS_VER }}\
         ${{ needs.compute-matrix.outputs.ALPHA_TAG }}-\
         cuda${{ matrix.CUDA_VER }}-\
-        py${{ matrix.PYTHON_VER }}"
+        py${{ matrix.PYTHON_VER }}-\
+        ${{ matrix.ARCH }}"
   delete-temp-images:
     if: ${{ !cancelled() && needs.test.result == 'success' }}
-    needs: [compute-matrix, build-multiarch-manifest, test]
+    needs: [compute-matrix, build, test]
```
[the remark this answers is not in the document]
jameslamb: I can help explain.

`build` publishes images with tags like this:

```text
rapidsai/staging:docker-notebooks-706-24.10a-cuda12.0-py3.11-arm64
rapidsai/staging:docker-notebooks-706-24.10a-cuda12.0-py3.11-amd64
```

The `build-multiarch-manifest` job pushes a new manifest to DockerHub which combines those 2 images together, so that something like this will work:

```shell
docker pull \
    rapidsai/staging:docker-notebooks-706-24.10a-cuda12.0-py3.11
```

*(notice .... no `-{arch}` suffix)*

As of #702, that `build-multiarch-manifest` job is not run on pull requests at all.

https://github.com/rapidsai/docker/blob/1c27d9245fd9d99ee35981b970acaf10961ca45b/.github/workflows/build-test-publish-images.yml#L171-L172

I believe that was intentional, as a way to cut down on CI time and network calls. The combined manifests are a convenience for downstream consumers of these images, and not necessary for one CI job here to read the outputs from a prior job.

.github/workflows/build-test-publish-images.yml
jameslamb: This version only affects `pre-commit` checking of the glue code in this repo, like https://github.com/rapidsai/docker/blob/branch-24.10/context/test_notebooks.py ... not any Python code that ever imports anything from RAPIDS.

And it's only in the diff here because I'm moving things around.

But while we're now talking about it... I think it would be safe (and one less source of maintenance burden) to remove this `setup-python` entirely and just take whatever version ships on the `ubuntu-latest` VM image from GitHub Actions.

.github/workflows/build-test-publish-images.yml
jameslamb: No, it (I think intentionally) does not run on PRs: https://github.com/rapidsai/docker/pull/708#discussion_r1718693278

jameslamb: @AyodeAwe what was the reason for not running `build-multiarch-manifest` on PRs any more as of #702?

I made some assumptions here but maybe those assumptions are wrong.

.github/workflows/build-test-publish-images.yml
raydouglass, line 257: Won't this change mean that for branch builds `delete-temp-images` will not run at all?

jameslamb: I don't think so? Because this workflow is invoked on branch builds.

https://github.com/rapidsai/docker/blob/1c27d9245fd9d99ee35981b970acaf10961ca45b/.github/workflows/publish.yml#L3-L6

https://github.com/rapidsai/docker/blob/1c27d9245fd9d99ee35981b970acaf10961ca45b/.github/workflows/publish.yml#L21-L26

 .... BUT it looks like `delete-temp-images` is ALREADY not running on branch builds 🙃 

Look at the most recent one:

https://github.com/rapidsai/docker/actions/runs/10405157692

Screenshot 2024-08-15 at 12 19 51 PM

And that's because it has `[test]` in `needs:` but tests are being skipped, because on branch builds this is run like this:

```text
build_type: branch
run_tests: false
```

jameslamb: These requirements can't all be satisfied:

* "do not run `test` on branch builds"
* "only run `delete-temp-images` after the tests pass"
* "always run `delete-temp-images` on branch builds"

jameslamb: @raydouglass what do you think about just completely eliminating `delete-temp-images` and instead relying on the other, higher-level cleanup of the `rapidsai/staging` DockerHub repo: https://github.com/rapidsai/workflows/blob/main/.github/workflows/cleanup_staging.yaml

That'd simplify things here a lot, I think.

raydouglass: Not necessarily opposed to letting the cleanup happen there, but what is the condition was updated to be "only run delete-temp-images after the test succeeds or is skipped"?

`if: ${{ !cancelled() && (needs.test.result == 'success'  || needs.test.result == 'skipped') }}` maybe?

.github/workflows/build-test-publish-images.yml
AyodeAwe: `build-multiarch-manifest` was never usually run in PRs:

https://github.com/rapidsai/docker/blob/5cea518e52c6e795557a2f23ffa8320ce38cc783/.github/workflows/build-test-publish-images.yml#L194

.github/workflows/build-test-publish-images.yml
jakirkham: > And it's only in the diff here because I'm moving things around.

Yeah was wondering for this reason whether this should be its own issue for separate follow up. That said, no strong feelings either way

.github/workflows/build-test-publish-images.yml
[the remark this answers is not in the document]
jakirkham: Thanks James! 🙏

That's a helpful explanation. Agree we don't want this constraint in PRs

Though I think with builds on branches, this was a change that Jake introduced in PR ( https://github.com/rapidsai/docker/pull/702 ) to make sure we were not deleting images before the manifest was generated

https://github.com/rapidsai/docker/blob/1c27d9245fd9d99ee35981b970acaf10961ca45b/.github/workflows/build-test-publish-images.yml#L232-L234

Otherwise if a build failed, we would delete the images needed for the manifest and restarting the failed jobs would just fail immediately (as the manifest cannot be created)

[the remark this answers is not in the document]
jameslamb: gah you're 100% right. We need to account for that here. I'll push some changes.

That condition never could have worked on PRs, because `build-multiarch-manifest` doesn't run on PRs (see related conversation in this other thread: https://github.com/rapidsai/docker/pull/708#discussion_r1718781188). Sorry I didn't catch it in previous reviews.

.github/workflows/build-test-publish-images.yml
jameslamb: Ah! Ok great, thanks for checking my understanding there.

.github/workflows/build-test-publish-images.yml
jameslamb: Ok sure, I just pushed a commit with this modified condition.

I do strongly support just dropping `delete-temp-images` entirely. If you'd prefer to discuss that separately from this PR let me know, I can write up an issue.

I'd make CI here work more like it does in other RAPIDS repos, which might reduce the incidence of bugs like those being fixed in this PR.

It'd also remove some complexity... no need to keep https://github.com/rapidsai/docker/blob/branch-24.10/ci/delete-temp-images.sh up to date with changes to the set of image tags or changes in DockerHub's API.

It'd also remove a source of network calls in CI here, which reduces the risk of networking-based job failures.

raydouglass (approved):

.github/workflows/build-test-publish-images.yml
jameslamb: I pushed a commit just removing `setup-python` and re-using whatever Python and `pip` is around in the `ubuntu-latest` image from GitHub Actions.

Seems like it worked well: https://github.com/rapidsai/docker/actions/runs/10408813338/job/28827015511?pr=708

I think we should do that until we encounter a version-related issue. The set of `pre-commit` checks here is pretty minimal and only running on a bit of glue code... I think the risk of issues is pretty low, and worth it in exchange for having one less place to update as our Python support matrix shifts.

.github/workflows/build-test-publish-images.yml
jameslamb: I just pushed a commit making `build-multiarch-manifest` run on PRs.

I think it just makes everything simpler...:

* it can now be required by the `pr-builder` job (to be sure it isn't accidentally skipped)
* issues with it can be detected on PRs, instead of only on branch builds
* it can be unconditionally included in the `needs:` block for `delete-temp-images`

Those jobs tend to take like 5-30 seconds each and don't use GPUs. I think that's a pretty small price to pay in exchange for the benefits I listed above.

jameslamb:
This is ready to merge, as soon as the branch protections are changed as described in https://github.com/rapidsai/docker/pull/708#issuecomment-2291493969

jameslamb:
Awesome, thanks so much for the help!

PR #717, opened 2024-10-21
Add support for Ubuntu 24.04.

bdice:
Part of https://github.com/rapidsai/build-planning/issues/74.

### DIFF
--- a/ci/compute-matrix.jq
+++ b/ci/compute-matrix.jq
@@ -3,6 +3,10 @@ def compute_arch($x):
 
 def compute_ubuntu_version($x):
   if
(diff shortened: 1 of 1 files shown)


jameslamb (approved):

jakirkham:
Not ideal, but let's try the old close and reopen. Especially now that CI will be less packed

bdice:
This was failing to merge because of a branch protection on `pr-builder / run` when the job is actually titled `ci / docker / pr-builder / run`. This branch protection was automatically added during branch-24.12 creation. We will need to fix this in our release automation scripts.

PR #723, opened 2024-12-17
Use patch file to use requests to download dataset for 24.12

dantegd:
This PR is to work around the 403 issue being seen when downloading datasets, while we create a more permanent solution.

### DIFF
--- a/context/cuvs-bench/cuvs_bench_get_dataset.patch
+++ b/context/cuvs-bench/cuvs_bench_get_dataset.patch
@@ -8,8 +8,8 @@ index a6b154ef..b023fcbd 100644
  import sys
 -from urllib.request import urlretrieve
 +import requests
- 
- 
+
+
  def get_dataset_path(name, ann_bench_data_path):
 @@ -29,7 +29,12 @@ def get_dataset_path(name, ann_bench_data_path):
  def download_dataset(url, path):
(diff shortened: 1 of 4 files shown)


context/cuvs-bench/cuvs_bench_get_dataset.patch
raydouglass, line 1: If going this route, this patch file needs to be copied into the container, like here: https://github.com/rapidsai/docker/blob/branch-25.02/cuvs-bench/gpu/Dockerfile#L55

context/cuvs-bench/cuvs_bench_get_dataset.patch
```diff
@@ -0,0 +1,27 @@
+diff --git a/python/cuvs_bench/cuvs_bench/get_dataset/__main__.py b/python/cuvs_bench/cuvs_bench/get_dataset/__main__.py
+index a6b154ef..b023fcbd 100644
+--- a/python/cuvs_bench/cuvs_bench/get_dataset/__main__.py
++++ b/python/cuvs_bench/cuvs_bench/get_dataset/__main__.py
+@@ -17,7 +17,7 @@ import argparse
+ import os
+ import subprocess
+ import sys
+-from urllib.request import urlretrieve
++import requests
```
raydouglass, line 10: Does the container have `requests` installed?

raydouglass (approved):

bdice (commented):
Why patch this if we can make the fix directly in cuvs-bench?

edit: Got the answer offline. 24.12 cuvs-bench artifacts were already published. This change could be made for 25.02, but we are going to try and self-host the datasets to avoid this problem entirely.

PR #725, opened 2025-01-14
Separate building RAPIDS & cuVS images

raydouglass:
Separates the workflows for building the RAPIDS end user images and the cuVS images.

The cuVS images do not depend on the RAPIDS end user images, so they can be built in parallel. This also allows for finer grained retries in case of failures.

### DIFF
--- a/.github/workflows/build-cuvs-image.yml
+++ b/.github/workflows/build-cuvs-image.yml
@@ -24,12 +24,6 @@ on:
       RAPIDS_VER:
         required: true
         type: string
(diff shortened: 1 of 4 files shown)


raydouglass:
The test failures should be unrelated and are probably new changes in 25.02 that have to be addressed in `cuspatial`.

AyodeAwe (approved):
LGTM

jakirkham:
Thanks Ray and Jake! 🙏

PR #742, opened 2025-03-09
Remove cuSpatial notebooks

bdice:
This PR removes cuSpatial notebooks from the notebook container. Most of these notebooks are unsupported for various reasons, so this ensures the notebook container ships examples that are widely supported.

### DIFF
--- a/context/notebooks.sh
+++ b/context/notebooks.sh
@@ -8,7 +8,7 @@
 
 set -euo pipefail
 
-NOTEBOOK_REPOS=(cudf cuml cugraph cuspatial)
+NOTEBOOK_REPOS=(cudf cuml cugraph)
 
 mkdir -p /notebooks /dependencies
 for REPO in "${NOTEBOOK_REPOS[@]}"; do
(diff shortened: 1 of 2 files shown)


bdice:
Notebook failures include:
- a cuGraph failure like `No module named 'cugraph.experimental.datasets` from the Force Atlas notebook. I'm pretty sure this is a known failure? I've certainly seen it before. I'm not sure if it has been worked on yet.
- ARM jobs are also showing a Jitify failure, which we've been observing upstream in cuDF as well. We don't have a solution for this yet.

Given that these are known failures, I think this PR succeeded in its goals of removing cuSpatial notebooks.

gforsyth:
> * No module named 'cugraph.experimental.datasets

I believe this has been moved to `cugraph.datasets`

bdice:
Fix for that is here: https://github.com/rapidsai/cugraph/pull/4964

jameslamb (approved):
Great, thank you for doing this.

raydouglass (approved):

Issue #667, opened 2024-05-01
[FEA] add tests on image characteristics


jameslamb:
**Is your feature request related to a problem? Please describe.**

Images published from this repo are intended for use in a wide variety of contexts. For example, they're used in several different cloud providers' bring-your-own container solutions (see https://docs.rapids.ai/deployment/stable/).

Those environments each have slightly different requirements for images you bring to them... things like default user, exposed ports, expected filepaths, etc.

When changes like #449 are made to modify the images here, there are currently not strong automatic protections against accidental regressions.

This issue proposes adding such protections, to improve release confidence and hopefully catch some types of issues during development instead of further downstream.

**Describe the solution you'd like**

I'm proposing the following:

* on every commit to PRs or `main` here, test all built images with `container-canary` (https://github.com/NVIDIA/container-canary), with the following tests:
   - compliance with various target deployment environments (with tests sourced from https://github.com/NVIDIA/container-canary/tree/main/examples)
   - other RAPIDS-specific tests to prevent accidental regressions (with tests checked into source control here)
* on every commit to PRs, test files in this repo with some static analyzers. For example:
   - `hadolint` for Dockerfiles ([link](https://github.com/hadolint/hadolint))
   - `shellcheck` for shell scripts ([link](https://github.com/koalaman/shellcheck))
   - `ruff` for Python code ([link](https://github.com/astral-sh/ruff))

**Describe alternatives you've considered**

N/A

**Additional context**

This proposal comes out of mine and @jacobtomlinson 's discussions around this: https://github.com/rapidsai/deployment/issues/277

And would help with stuff like this:

* https://github.com/rapidsai/docker/issues/582

I'd be happy to come do the work, just looking for a 👍🏻 / 👎🏻 on the general idea (cc @raydouglass @ajschmidt8 @AyodeAwe )

jameslamb:
Based on my experiments in #670 so far, I think this will require some updates to `container-canary`:

* [ ] https://github.com/NVIDIA/container-canary/issues/61
* [ ] https://github.com/NVIDIA/container-canary/issues/62

I'd like to try contributing those, at some point.

jameslamb:
Some good discussion on possible `container-canary` tests here:

* https://github.com/rapidsai/docker/issues/712#issuecomment-2386370049
* https://github.com/rapidsai/docker/issues/712#issuecomment-2388384139

jameslamb:
Once #747 is complete, I think we can close this issue.

[...]

PR #747, opened 2025-04-18
lint Dockerfiles with 'hadolint'

jameslamb:
Contributes to #667

Proposes running `hadolint` (https://github.com/hadolint/hadolint) on Dockerfiles in this repo. I've found that this tool makes some of the same recommendations that I find myself making in PR reviews... I hope enforcing it in CI will reduce reviewing effort a bit and improve our release confidence.

All the Dockerfiles contain changes for these reported issues, which I agree with:

```text
DL3015 info: Avoid additional packages by specifying `--no-install-recommends`
DL3047 info: Avoid use of wget without progress bar. Use `wget --progress=dot:giga <url>`. Or consider using `-q` or `-nv` (shorthands for `--quiet` or `--no-verbose`).
```

See other comments inline.

### Why not use `pre-commit`?

As @KyleFromNVIDIA pointed out to me a while ago, the install hooks for `hadolint`'s `pre-commit` hook are broken: https://github.com/hadolint/hadolint/issues/886

We don't change these Dockefiles **that** often... I think the slight extra friction of having to run a script manually or just get feedback from CI should not block us from adopting this tool.

### DIFF
--- a/.github/workflows/build-test-publish-images.yml
+++ b/.github/workflows/build-test-publish-images.yml
@@ -50,6 +50,9 @@ jobs:
       - uses: actions/checkout@v4
       - name: Run pre-commit
         uses: pre-commit/action@v3.0.1
+      - name: Run hadolint
+        run: |
+          ci/lint-dockerfiles.sh
   compute-matrix:
     runs-on: ubuntu-latest
     container:
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,11 @@ ARG LINUX_VER=${LINUX_DISTRO}${LINUX_DISTRO_VER}
 ARG RAPIDS_VER=25.06
 
 # Gather dependency information
+
+# ignore hadolint DL3007... we really do always want the latest 'rapidsai/ci-conda',
+# and don't want to have to push new commits to update to it
+#
+# hadolint ignore=DL3007
 FROM rapidsai/ci-conda:latest AS dependencies
 ARG CUDA_VER
 ARG PYTHON_VER
@@ -27,7 +32,7 @@ COPY notebooks.sh /notebooks.sh
 
 RUN <<EOF
 apt-get update
-apt-get install -y rsync
+apt-get install -y --no-install-recommends rsync
 /notebooks.sh
 apt-get purge -y --auto-remove rsync
 rm -rf /var/lib/apt/lists/*
@@ -45,11 +50,12 @@ SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
 RUN <<EOF
 apt-get update
-apt-get install -y wget
-wget https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - | tar -xz -C /usr/local/bin
+apt-get install -y --no-install-recommends wget
+wget --quiet https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - | tar -xz -C /usr/local/bin
 apt-get purge -y --auto-remove wget
 rm -rf /var/lib/apt/lists/*
 EOF
+
 RUN useradd -rm -d /home/rapids -s /bin/bash -g conda -u 1001 rapids
 
 USER rapids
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -45,11 +50,12 @@ SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
...
 RUN <<EOF
 apt-get update
-apt-get install -y wget
-wget https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - | tar -xz -C /usr/local/bin
+apt-get install -y --no-install-recommends wget
+wget --quiet https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - | tar -xz -C /usr/local/bin
 apt-get purge -y --auto-remove wget
 rm -rf /var/lib/apt/lists/*
 EOF
+
```
jameslamb, line 58: This is necessary. `hadolint` errors on processing a Dockerfile that has anything other than a blank line after the end of a heredoc: https://github.com/hadolint/hadolint/issues/960

Without this change, this fails like this:

```shell
ci/lint-dockerfiles.sh
# linting './Dockerfile'
# /dev/stdin:58:1 unexpected 'R' expecting a new line followed by the next instruction
```

cuvs-bench/cpu/Dockerfile
```diff
@@ -55,8 +55,7 @@ CMD ["--dataset fashion-mnist-784-euclidean", "", "--algorithms hnswlib"]
 
 ENTRYPOINT ["/bin/bash", "/data/scripts/run_benchmark.sh"]
 
-
-FROM bench-base AS cuvs-bench-cpu-datasets
+FROM cuvs-bench-cpu AS cuvs-bench-cpu-datasets
```
jameslamb, line 58: This is a serious bug in this Dockerfile!

And `hadolint` caught it (indirectly) 😁 

```text
DL3006 warning: Always tag the version of an image explicitly
```

There's not an earlier target called `bench-base`, so trying to build this would try to pull an image from DockerHub called `bench-base`. I'm pretty confident this is not what we want.

jameslamb:
Ci will be blocked until #749 is resolved, but I think this is still ready to review.

.github/workflows/build-test-publish-images.yml
```diff
@@ -50,6 +50,9 @@ jobs:
       - uses: actions/checkout@v4
       - name: Run pre-commit
         uses: pre-commit/action@v3.0.1
+      - name: Run hadolint
+        run: |
+          ci/lint-dockerfiles.sh
```
jacobtomlinson, line 55: It looks like `hadolint` has pre-commit hooks. I'm curious why you're not using them directly?

https://github.com/hadolint/hadolint/blob/master/.pre-commit-hooks.yaml

jacobtomlinson (commented):
Nice tool!

.github/workflows/build-test-publish-images.yml
jameslamb: I mentioned this directly in the PR description.

> ### Why not use `pre-commit`?
>
> _As @KyleFromNVIDIA pointed out to me a while ago, the install hooks for `hadolint`'s `pre-commit` hook are broken: https://github.com/hadolint/hadolint/issues/886_
>
> _We don't change these Dockefiles **that** often... I think the slight extra friction of having to run a script manually or just get feedback from CI should not block us from adopting this tool._

Because of that issue, you'd have to manually install `hadolint` to run `pre-commit run` locally... I don't think the friction is worth it.

jacobtomlinson: Gah sorry I totally missed that section. Makes sense!

jameslamb: ha no problem

raydouglass (approved):

jameslamb:
Thanks, friend!

Realized I'd left some commented-out configuration behind, just pushed https://github.com/rapidsai/docker/pull/747/commits/af3470ff2db9261b4818b6314b26dc58e7d96349 to remove that.

Once CI passes, I'll merge this.

PR #750, opened 2025-04-28
Install curl

raydouglass:
Some users utilize curl across many images to perform healthchecks. In `25.04`, the `cuspatial` notebooks were dropped which also dropped the image's dependency on `curl`.

So this PR explicitly installs the `curl` system package. And just to simplify things, switches to use `curl` to download instead of `wget`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,9 +50,8 @@ SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
 RUN <<EOF
 apt-get update
-apt-get install -y --no-install-recommends wget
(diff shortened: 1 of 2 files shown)


bdice (approved):

jameslamb (approved):
Don't hold up the PR over this comment.... but I think it should be made clearer that we want to include `curl` as part of the user-facing guarantees about these images, not just as a tool to download the `gha-tools` scripts here in the `docker` build.

That makes it a good candidate for a `container-canary` test (https://github.com/rapidsai/docker/blob/branch-25.06/tests/container-canary/base.yml), or at least a code comment in `Dockerfile` to prevent removing it in the future.

PR #763, opened 2025-05-28
Add compilers to base & notebooks images

raydouglass:
Some use cases of these images install a few pip packages which require building. So this PR ensures that the compilers are available.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,6 +77,7 @@ rapids-mamba-retry install -y -n base \
     "rapids=${RAPIDS_VER}.*" \
     "python=${PYTHON_VER}.*" \
     "cuda-version=${CUDA_VER%.*}.*" \
+    compilers \
     ipython
 conda clean -afy
 EOF
--- a/tests/container-canary/base.yml
+++ b/tests/container-canary/base.yml
@@ -65,6 +65,29 @@ checks:
         command:
           - wget
           - --version
+  # some use-cases install new packages using 'pip', requiring a build
+  # So check that compilers are available
+  - name: tool-gcc
+    description: gcc can be executed
+    probe:
+      exec:
+        command:
+          - gcc
+          - --version
+  - name: tool-g++
+    description: g++ can be executed
+    probe:
+      exec:
+        command:
+          - g++
+          - --version
+  - name: tool-gfortran
+    description: gfortran can be executed
+    probe:
+      exec:
+        command:
+          - gfortran
+          - --version
   - name: user-is-rapids
     description: Default user is rapids (uid=1001)
     probe:
--- a/tests/container-canary/notebooks.yml
+++ b/tests/container-canary/notebooks.yml
@@ -15,3 +15,26 @@ checks:
           - jupyter
           - lab
           - --version
+  # some use-cases install new packages using 'pip', requiring a build
+  # So check that compilers are available
+  - name: tool-gcc
+    description: gcc can be executed
+    probe:
+      exec:
+        command:
+          - gcc
+          - --version
+  - name: tool-g++
+    description: g++ can be executed
+    probe:
+      exec:
+        command:
+          - g++
+          - --version
+  - name: tool-gfortran
+    description: gfortran can be executed
+    probe:
+      exec:
+        command:
+          - gfortran
+          - --version


taureandyernv:
Sure, @jameslamb.  For example, today, we support Digital Biology, and through them SCVerse's RAPIDS-singlecell library.  They use scikit-misc to as a dependency for install.  Previously, up til 25.04, those listed compilers were in our docker container and the pip install would work.  Today, in 25.06, we found that those compilers are missing, causing their install to fail, and breaking our ability to support them in our containers for this release and beyond.  This fix will restore our ability to support them and their installs, and other similar installs from around the community.

jameslamb:
> _up til 25.04, those listed compilers were in our docker container and the pip install would work_

That doesn't seem true to me:

```shell
docker run \
  --rm \
  -it rapidsai/base:24.12-cuda12.5-py3.10 \
  /bin/bash -c "gcc --version"
# /bin/bash: line 1: gcc: command not found

docker run \
  --rm \
  -it rapidsai/base:25.02-cuda12.8-py3.10 \
  /bin/bash -c "gcc --version"
# /bin/bash: line 1: gcc: command not found

docker run \
  --rm \
  -it rapidsai/base:25.04-cuda12.8-py3.10 \
  /bin/bash -c "gcc --version"
# /bin/bash: line 1: gcc: command not found

docker run \
  --rm \
  -it rapidsai/base:25.06a-cuda12.8-py3.10 \
  /bin/bash -c "gcc --version"
# /bin/bash: line 1: gcc: command not found
```

But without **specifics** about exactly the images you're using and how you're running them, it's hard to know if what I've tried above is even related to what you're reporting. For example I don't know if "our docker container" refers to the `rapidsai/base` images or something else.

Could you please provide (or ask whoever you're working with to provide) some specific evidence and description of the problem, in the form of a `docker run` command that fails and which you think should succeed?

For example... maybe it's actually that whatever `pip install` you're doing is trying to install `scikit-misc` for a combination of CPU architecture, Python version, and operating system that it doesn't have wheels for, and maybe earlier versions did. If that's the case, maybe the fix for your case is to pass `--prefer-binary` to `pip install`.

Or maybe it is architecture dependent? I see that the latest release of `scikit-misc` does not have `aarch64` wheels: https://pypi.org/project/scikit-misc/0.5.1/#files

And so on `aarch64` systems, this fails:

```shell
docker run \
  --rm \
  -it rapidsai/base:25.06a-cuda12.8-py3.10 \
  /bin/bash -c "pip install --prefer-binary 'scikit-misc'"
```

Like this:

```text
Collecting scikit-misc
  Downloading scikit_misc-0.5.1.tar.gz (298 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [20 lines of output]
      + meson setup /tmp/pip-install-buxh_q95/scikit-misc_34f755f3b56e4737bba8e9614c885f70 /tmp/pip-install-buxh_q95/scikit-misc_34f755f3b56e4737bba8e9614c885f70/.mesonpy-c1cbxh6n -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md --native-file=/tmp/pip-install-buxh_q95/scikit-misc_34f755f3b56e4737bba8e9614c885f70/.mesonpy-c1cbxh6n/meson-python-native-file.ini
      The Meson build system
      Version: 1.8.1
      Source dir: /tmp/pip-install-buxh_q95/scikit-misc_34f755f3b56e4737bba8e9614c885f70
      Build dir: /tmp/pip-install-buxh_q95/scikit-misc_34f755f3b56e4737bba8e9614c885f70/.mesonpy-c1cbxh6n
      Build type: native build
      Project name: scikit-misc
      Project version: 0.5.1

      ../meson.build:2:0: ERROR: Unknown compiler(s): [['cc'], ['gcc'], ['clang'], ['nvc'], ['pgcc'], ['icc'], ['icx']]
      The following exception(s) were encountered:
      Running `cc --version` gave "[Errno 2] No such file or directory: 'cc'"
      Running `gcc --version` gave "[Errno 2] No such file or directory: 'gcc'"
      Running `clang --version` gave "[Errno 2] No such file or directory: 'clang'"
      Running `nvc --version` gave "[Errno 2] No such file or directory: 'nvc'"
      Running `pgcc --version` gave "[Errno 2] No such file or directory: 'pgcc'"
      Running `icc --version` gave "[Errno 2] No such file or directory: 'icc'"
      Running `icx --version` gave "[Errno 2] No such file or directory: 'icx'"

      A full log can be found at /tmp/pip-install-buxh_q95/scikit-misc_34f755f3b56e4737bba8e9614c885f70/.mesonpy-c1cbxh6n/meson-logs/meson-log.txt
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed
```

But on an x86_64 system, I suspect that would succeed (hitting `send` now without testing because my train is about to pull in).

In that case, a better solution might be to work with `scikit-misc` to get aarch64 packages published.

I'm sorry, but I'm really not comfortable growing these already-6GB images even more without a more specific description of the problem we're trying to fix.

raydouglass:
> They use scikit-misc to as a dependency for install. Previously, up til 25.04, those listed compilers were in our docker container and the pip install would work. Today, in 25.06, we found that those compilers are missing, causing their install to fail

It sounds like they're using our docker images in a way that's not really the intention. These images are for end-users to run RAPIDS and, for convenience, we have a feature to install packages at runtime. I don't think these images were intended to support more than that.

The use case describes sounds more like they should be building a new image, possibly on top of ours, though that's still not really our intention. They would install the packages (compilers, `scikit-misc`, etc) they need during build time. This would definitely be the more docker-esque approach.

raydouglass:
Closing for now at least. Will revisit in `25.08`.

PR #768, opened 2025-06-10
Remove dask, pytorch, and nvidia channels

vyasr:
Contributes to https://github.com/rapidsai/build-planning/issues/166

### DIFF
--- a/.github/workflows/build-cuvs-image.yml
+++ b/.github/workflows/build-cuvs-image.yml
@@ -65,7 +65,7 @@ jobs:
           GIT_DESCRIBE_TAG="${GIT_DESCRIBE_TAG:1}" # remove leading 'v'
           if [[ ! $GIT_DESCRIBE_TAG =~ [a-z] ]]; then
             rapids-logger "Most recent tag is for release, adding the `rapidsai` channel and removing the `rapidsai-nightly` channel."
-            sed -i 's|rapidsai-nightly|rapidsai|d' context/condarc
+            sed -i 's|rapidsai-nightly|rapidsai|' context/condarc
           else
             rapids-logger "Most recent tag is an alpha. Build will use nightly channels."
           fi
--- a/.github/workflows/build-cuvs-image.yml
+++ b/.github/workflows/build-cuvs-image.yml
@@ -64,8 +64,8 @@ jobs:
           GIT_DESCRIBE_TAG="$(git describe --tags --abbrev=0)"
           GIT_DESCRIBE_TAG="${GIT_DESCRIBE_TAG:1}" # remove leading 'v'
           if [[ ! $GIT_DESCRIBE_TAG =~ [a-z] ]]; then
-            rapids-logger "Most recent tag is for release, adding `rapidsai` channel and removing `rapidsai-nightly` and `dask/label/dev` channels."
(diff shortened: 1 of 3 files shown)


.github/workflows/build-cuvs-image.yml
```diff
@@ -64,8 +64,8 @@ jobs:
           GIT_DESCRIBE_TAG="$(git describe --tags --abbrev=0)"
           GIT_DESCRIBE_TAG="${GIT_DESCRIBE_TAG:1}" # remove leading 'v'
           if [[ ! $GIT_DESCRIBE_TAG =~ [a-z] ]]; then
-            rapids-logger "Most recent tag is for release, adding `rapidsai` channel and removing `rapidsai-nightly` and `dask/label/dev` channels."
-            sed -i 's|rapidsai-nightly|rapidsai|;\|dask/label/dev|d' context/condarc
+            rapids-logger "Most recent tag is for release, adding the `rapidsai` channel and removing the `rapidsai-nightly` channel."
+            sed -i 's|rapidsai-nightly|rapidsai|d' context/condarc
```
jameslamb, line 68:
```suggestion
            sed -i 's|rapidsai-nightly|rapidsai|' context/condarc
```

I *think* the `|d` was there to delete `dask/label/dev`. Now that this is only "replace `rapidsai-nightly` with `rapidsai`", we want to omit it.

```shell
cat > condarc <<EOF
- rapidsai-nightly
EOF

sed -i 's|rapidsai-nightly|rapidsai|d' ./condarc
# sed: -e expression #1, char 29: unknown option to `s'

sed -i 's|rapidsai-nightly|rapidsai|' ./condarc
cat ./condarc
# - rapidsai
```

jameslamb (commented):
Support this change generally, but I think that `sed` change needs to look a little different.

context/condarc
```diff
@@ -1,7 +1,6 @@
 auto_update_conda: False
 channels:
   - rapidsai-nightly
-  - dask/label/dev
   - pytorch
   - conda-forge
   - nvidia
```
bdice, line 6: Please also remove `nvidia` here. And do we need `pytorch` anymore? They aren't shipping conda packages there anymore.

bdice (changes requested):

context/condarc
vyasr: Sure I can expand the scope of this one.

vyasr:
Well that's problematic. It means we've been pulling invalid environments so far. cufile packages didn't exist on arm until 12.2, so if we were getting them from nvidia (likely the correct diagnosis) we were getting packages that weren't actually constrained to the right cuda-version. I wonder why we have these though; our [cufile dependency should be conditional on the cuda version](https://github.com/rapidsai/cudf/blob/branch-25.08/conda/recipes/libcudf/recipe.yaml#L118).

vyasr:
Ah the problem is kvikio's recipe, not cudf's. I have to look a bit closer to figure out the correct fix here though; kvikio has a tighter connection to cufile than cudf, so I don't know if simply removing the dependency is the right answer here.

vyasr:
OK I've put in a temporary patch in https://github.com/rapidsai/kvikio/pull/753 to get around this issue. We'll need a follow up that sets up the correct constraints, probably with a variant build.

jameslamb (approved):
🎉

PR #771, opened 2025-06-24
use docker cache, run pre-commit updates less frequently

jameslamb:
Replaces #770

Proposes 2 small changes to help with development here:

* using the DockerHub pull-through cache on NVIDIA's self-hosted runners
  - *to reduce the disruption from network issues and rate limits (see https://github.com/rapidsai/ci-imgs/pull/281#issuecomment-3001418760)*
* switching the `pre-commit.ci` autoupdates from weekly to monthly
  - *this project uses some tools like `ruff` that update very frequently, meaning we get a new PR from `pre-commit.ci` almost every week... CI here uses a LOT of runners, and for this repo I don't think that resource usage is worth it in exchange for moving to a new version of `ruff` a few weeks earlier than we otherwise would*

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,7 @@
 # Copyright (c) 2025, NVIDIA CORPORATION.
 ---
+ci:
+  autoupdate_schedule: 'monthly'
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
@@ -7,7 +9,7 @@ repos:
       - id: end-of-file-fixer
       - id: trailing-whitespace
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.13
+    rev: v0.12.0
     hooks:
       - id: ruff
         args: ["--config", "pyproject.toml"]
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -1,13 +1,15 @@
 # Copyright (c) 2025, NVIDIA CORPORATION.
 ---
+ci:
+  autoupdate_schedule: 'monthly'
```
jameslamb, line 4: docs on this: https://pre-commit.ci/#configuration

jameslamb:
Tests here should pass once https://github.com/rapidsai/cuml/pull/6922 is merged.

Otherwise, I think this is ready for review.

jameslamb:
Thanks for the re-run @msarahan ! In theory https://github.com/rapidsai/cuml/pull/6922 should fix this.

Looking at the re-run build jobs though, seems that they got slightly old nightlies... `cuml 25.08.00a169` (https://github.com/rapidsai/docker/actions/runs/15858877647/job/44801626136?pr=771#step:9:649), when the latest is `25.08.00a174` (https://anaconda.org/rapidsai-nightly/cuml).

Going to push a fresh, empty commit to fully re-trigger everything. Hopefully that + the fact that the build job from that `cuml` commit finishing (https://github.com/rapidsai/cuml/actions/runs/15883547057/job/44789783760) will be enough to fix this.

PR #823, opened 2025-12-03
renovate: group GitHub Actions updates into fewer PRs

jameslamb:
Proposes updating the `renovate` configuration so that almost all "hey there's a new version of this third-party GitHub Action" updates are grouped together into a single PR.

To avoid cases like this where multiple individual PRs are opened at the beginning of the month:

* #808
* #807
* #806

This should reduce CI usage, notifications, and review burden a bit.

### How I made this list

```shell
git grep -E 'uses\: '
```

### How to test this

I'm not sure how to test this without merging it.

In https://github.com/rapidsai/ci-imgs/pull/265, we just merged a similar change and then looked at the dashboard issue to check.

So I think we should just merge this and then check https://github.com/rapidsai/docker/issues/645

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -11,6 +11,12 @@
         "^(actions/.*|aws-actions/*|docker/*|go|peter-evans/dockerhub-description|pre-commit/)$"
       ],
       "groupName": "gh-actions"
+    },
+    {
+      "matchDatasources": ["docker"],
+      "matchPackagePatterns": [
+        "^condaforge/miniforge3$"
+      ]
     }
   ]
 }


gforsyth:
> Oh to be clear, my intention in this PR is to still keep `miniforge3` separate. I thought having a separate entry for it in ` "packageRules"` would accomplish that. Did I make a mistake there?

Honestly, I don't know.  I say we merge it and look at the dashboard and proceed from there.

jameslamb:
ayyyyyy nice! Thanks for the help.

jameslamb:
> _To get an update, click the checkbox below_

I did that and it created #825

Looks like it didn't include `aws-actions/*` because it didn't like the way I closed #807, and didn't update `go` because that particular update is rate limited (don't really understand what that means in this context but 🤷🏻).

image

But overall I'm feeling like this is working. Seeing those 3 things grouped together in #825 is a good sign.

PR #859, opened 2026-03-05
Test conda env activation in container-canary tests

jakirkham:
Fixes https://github.com/rapidsai/docker/issues/857

For scripts like `profile.sh` to run (needed for Conda activation) we need to use a login shell. Hence the addition of `-l` (L) here.

Also interactive shells tend to behave a bit better from a user experience perspective. So have added `-i` too.

### DIFF
--- a/.github/workflows/test-notebooks.yml
+++ b/.github/workflows/test-notebooks.yml
@@ -71,7 +71,7 @@ jobs:
     steps:
       - name: Install required tools
         run: |
-          mamba install -n base --freeze-installed \
+          conda install -n base --freeze-installed \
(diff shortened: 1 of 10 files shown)


jameslamb (changes requested):
Thanks for starting this, it's an interesting approach to the environment-activation problem. I'm leaving a blocking review because I have some concerns.

jameslamb (approved):
Thanks for the changes, I'm good with this as a way to help with the startup environment for interactive uses.

I think we will want both this AND #860 (which fixes the problem generally for non-interactive use too).

PR #867, opened 2026-03-30
ensure the latest tag is always used to determine version

jameslamb:
Nightly builds on `main` are failing like this:

```text
56.98 Could not solve for environment specs
56.98 The following packages are incompatible
56.98 ├─ rapids-dask-dependency =26.6,>=0.0.0a0 * is requested and can be installed;
56.98 └─ rapids =26.4 * is not installable because it requires
56.98    └─ custreamz =26.4 *, which requires
56.98       └─ rapids-dask-dependency =26.4 *, which conflicts with any installable versions previously reported.
```

([build link](https://github.com/rapidsai/docker/actions/runs/23732578503/job/69129525562#step:10:731))

They should be looking for 26.06 packages, and I'd expect them to because the `v26.06.00a` tag is closer to HEAD of `main` than `v26.04.00a`.

image

I think the root cause is that I made some mistake in creating the branch for #866 (still not sure exactly what) that ended up making `v26.04.00a` satisfy this:

https://github.com/rapidsai/docker/blob/2637556074567056aa69a7cb1850ac844ef89748/.github/workflows/build-cuvs-image.yml#L63

This proposes permanently avoiding by using `git describe --first-parent`

From https://git-scm.com/docs/git-describe

> _Follow only the first parent commit upon seeing a merge commit. This is useful when you wish to not match tags on branches merged in the history of the target commit._

### How I tested this

```console
$ git checkout main
$ git pull upstream main
$ git fetch upstream --tags
$ git describe --tags --abbrev=0
v26.04.00a
$ git describe --tags --first-parent --abbrev=0
v26.06.00a
```

### DIFF
--- a/.github/workflows/build-cuvs-image.yml
+++ b/.github/workflows/build-cuvs-image.yml
@@ -60,7 +60,7 @@ jobs:
           fetch-depth: 0
       - name: Clean up condarc for release builds
         run: |
-          GIT_DESCRIBE_TAG="$(git describe --tags --abbrev=0)"
+          GIT_DESCRIBE_TAG="$(git describe --tags --first-parent --abbrev=0)"
           GIT_DESCRIBE_TAG="${GIT_DESCRIBE_TAG:1}" # remove leading 'v'
           if [[ ! $GIT_DESCRIBE_TAG =~ [a-z] ]]; then
             echo 'Most recent tag is for release, adding the `rapidsai` channel and removing the `rapidsai-nightly` channel.'
--- a/.github/workflows/build-rapids-image.yml
+++ b/.github/workflows/build-rapids-image.yml
@@ -61,7 +61,7 @@ jobs:
           fetch-depth: 0
       - name: Clean up condarc for release builds
         run: |
-          GIT_DESCRIBE_TAG="$(git describe --tags --abbrev=0)"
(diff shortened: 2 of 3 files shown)


jameslamb:
Yeah this is working, seeing `RAPIDS_VER=26.06` in CI logs.

When this is approved, let's admin-merge past the failing `check-nightly-ci` check. This should fix the nightly builds.

jameslamb:
Tests are failing like this:

> ImportError: /opt/conda/bin/../lib/libcugraph_mg.so: undefined symbol: _ZN3rmm13device_bufferC1EmNS_16cuda_stream_viewENS_6detail23cccl_async_resource_refIN4cuda2mr3__412resource_refIJNS6_17device_accessibleEEEEEE

([build link](https://github.com/rapidsai/docker/actions/runs/23769371096/job/69259346588?pr=867))

That'll be fixed by a new build of cuGraph, once https://github.com/rapidsai/cugraph/pull/5479 is in

AyodeAwe (approved):

jameslamb:
Pushed an empty commit to trigger a new build, now that package builds (esp. for cuGraph) should be fixed.

jameslamb:
Everything but `check-nightly-ci` is now passing, admin-merging this. That should fix the nightly builds and get `check-nightly-ci` passing again.

PR #895, opened 2026-07-23
feat: publish signed image provenance manifests

msarahan:
**Posted by Codex (GPT-5.6) on behalf of Michael Sarahan. Treat this description as LLM-generated.**

## Summary

Add a signed, OCI-referrer RAPIDS image provenance manifest for published `base`
and `notebooks` image digests.

- Export exact `conda-meta` package facts through minimal BuildKit targets rather
  than downloading image layers.
- Attach a platform-specific manifest to every pushed architecture image, then
  attach a multiarch index manifest that links those platform records.
- Record the image digest, platform, source SHA, workflow URL, build arguments,
  and conda package name/version/build/channel/source URL.
- Keylessly sign each attached OCI artifact with GitHub Actions OIDC and Cosign.
- Keep pURLs explicitly unmapped until a verified upstream mapping is available;
  this intentionally does not invent `pkg:conda/...` identifiers.

## Relationship to #805

https://github.com/rapidsai/docker/pull/805 is complementary. It generates and
embeds a CycloneDX SBOM in the final image. This change makes digest-bound build
facts discoverable as registry metadata and supplies the platform-to-multiarch
relationship needed by scan triage without pulling image layers.

If #805 lands, the preferred follow-up is to add its SBOM digest/reference to
this manifest rather than maintain a separate provenance model. Its Dockerfile
stage renaming will require a small conflict resolution in this branch's
`provenance-base` and `provenance-notebooks` export targets.

## Validation

- `pytest -q tests/test_image_provenance_manifest.py`
- `pre-commit run --all-files` (Ruff, ShellCheck, Zizmor, copyright checks)
- YAML parsing and `git diff --check`

Buildx static validation was attempted but the local configured remote builder
timed out while connecting. No image was built locally. The first CI run should
therefore confirm Docker Hub OCI-referrer support and keyless Cosign signing on
the production runners.


msarahan:
superseded by #896 because CI can't run from my fork

